PR #2, opened 2016-06-27
fixes response overwrite bug

benlemasurier:
This fixes a bug in truncated response payload handling. Payloads requiring more than one call to `Read()` were overwiting the previous read buffer contents. Because we know the response payload size, we can simply call `io.ReadAll()` rather than looping over calls to `Read()`.


mdlayher:
:+1:

PR #4, opened 2016-07-19
check for QEMU response errors

benlemasurier:
When performing QEMU monitor commands, libvirt will return `StatusOK` even when the underlying QEMU process fails to perform the command.

This modifies `Run()` to check for QEMU errors.

I'm not entirely happy with the hacky modifications to the test library to handle this scenario. The test framework is in obvious need of a complete refactor. For now this will have to work.


kitschysynq:
lgtm

PR #7, opened 2016-08-19
Add support for domain migrations

benlemasurier:
This adds basic support for domain migrations from one hypervisor to another. Migration options, e.g., live, tunneled, compressed, etc.., are specified by the constants described `constants.Migrate*`.

Two unknowns remain, Libvirt specifies `RemoteParameters` and `CookieIn`. In testing both values are always set to 0 by `virsh` and the source does not provide clear definitions of their purpose. For now, using the same zero'd values as done by `virsh` will be Good Enough.


internal/constants/constants.go
```diff
@@ -53,3 +54,51 @@ const (
 	// UUIDSize is the length of a UUID, in bytes.
 	UUIDSize = 16
 )
+
+// domain migration flags
+const (
+	// MigrateLive performs a zero-downtime live migration
```
mdlayher: Add a period to each of these.

libvirt_test.go
```diff
@@ -43,6 +44,24 @@ func TestDisconnect(t *testing.T) {
...
+	var flags uint32
+	flags = constants.MigrateLive |
+		constants.MigratePeerToPeer |
+		constants.MigratePersistDestination |
+		constants.MigrateChangeProtection |
+		constants.MigrateAbortOnError |
+		constants.MigrateAutoConverge |
+		constants.MigrateNonSharedDisk
+
+	if err := l.Migrate("test", "qemu+tcp//foo/system", flags); err != nil {
```
mdlayher: Invalid URI here.  Should probably validate this in `Migrate` method too.

internal/constants/constants.go
```diff
@@ -53,3 +54,51 @@ const (
 	// UUIDSize is the length of a UUID, in bytes.
 	UUIDSize = 16
 )
+
+// domain migration flags
+const (
```
mdlayher: https://golang.org/pkg/net/#Flags

How about adding `type MigrateFlags uint32`, and then the below constants use that type and become `MigrateFlagsLive`, `MigrateFlagsPeerToPeer`, etc.

mdlayher: These can't live in `internal/constants/` because they need to be specified by the caller to `l.Migrate`.

libvirt.go
```diff
@@ -204,6 +204,48 @@ func (l *Libvirt) Events(dom string) (<-chan DomainEvent, error) {
...
+func (l *Libvirt) Migrate(dom string, dest string, flags uint32) error {
+	d, err := l.lookup(dom)
+	if err != nil {
+		return err
+	}
+
+	payload := struct {
+		Domain           Domain
+		DestinationURI   string
+		RemoteParameters uint32 // unused
```
mdlayher: > Two unknowns remain, Libvirt specifies RemoteParameters and CookieIn. In testing both values are always set to 0 by virsh and the source does not provide clear definitions of their purpose. For now, using the same zero'd values as done by virsh will be Good Enough.

Leave a short comment to this effect here for future reference.

libvirt.go
```diff
@@ -204,6 +204,48 @@ func (l *Libvirt) Events(dom string) (<-chan DomainEvent, error) {
 	return c, nil
 }
 
+// Migrate synchronously performs a domain migration to the hypervisor
+// specified by dest. The flags argument determines the type of migration
+// and how it will be performed. For more information on the available
+// migration flags and their meaning, see constants.Migrate*.
+func (l *Libvirt) Migrate(dom string, dest string, flags uint32) error {
```
mdlayher: Validate `dest` using `url.Parse`.

internal/constants/constants.go
benlemasurier: excellent suggestion! Done. However, I'm not sure what you mean by "These can't live in internal/constants/ because they need to be specified by the caller to l.Migrate."

libvirt.go
benlemasurier: :+1:

mdlayher:
https://github.com/digitalocean/go-libvirt/pull/7#discussion_r75533495

GitHub marked this as resolved, but we should still validate `dest` with `url.Parse`.

benlemasurier:
> GitHub marked this as resolved, but we should still validate dest with url.Parse.

saw that, done.

libvirt.go
```diff
@@ -204,6 +255,58 @@ func (l *Libvirt) Events(dom string) (<-chan DomainEvent, error) {
 	return c, nil
 }
 
+// Migrate synchronously performs a domain migration to the hypervisor
+// specified by dest. The flags argument determines the type of migration
```
mdlayher: I feel we should specify that `dom` is the name of the domain and that `dest` should be a URL probably of the form `qemu+tcp://` or `qemu+ssh://`, etc.  Leave examples.

libvirt.go
mdlayher: Add this comment and then I'm happy!

benlemasurier: :+1: , noted just above

mdlayher:
:+1:

PR #11, opened 2016-09-28
migrate: insert required request structure padding

benlemasurier:
Further testing reveals we need some additional padding on the request passed to libvirt when performing migrations. I believe this is due to the structure alignment normally provided automatically by the compiler. We've seen this in a few other areas within the package as well.

In testing, this fixes an error I was seeing: "Unable to decode message payload", returned from libvirt when performing migrations.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -277,12 +277,14 @@ func (l *Libvirt) Migrate(dom string, dest string, flags MigrateFlags) error {
 	// For now, using the same zero'd values as done by virsh will be Good Enough.
 	payload := struct {
 		Domain           Domain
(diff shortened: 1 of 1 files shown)


benlemasurier:
This has no effect on the existing tests for `Migrate()`, which is a good thing. We could modify `libvirttest/` to inspect the each request payload but that seems like overkill.

mdlayher:
Sounds like a larger chunk of work; perhaps worth breaking into its own issue if deemed necessary.

:+1:

benlemasurier:
Indeed. I'll revisit the issues we saw within travis for #12 before long.

PR #13, opened 2016-09-28
Add support for undefining a domain

benlemasurier:


libvirt.go
```diff
@@ -128,6 +128,20 @@ const (
 	MigrateFlagRDMAPinAll
 )
 
+// UndefineFlags specifies options available when undefining a domain.
+type UndefineFlags uint32
+
+const (
+	// UndefineFlagManagedSave removes all domain managed save data.
+	UndefineFlagManagedSave = 1 << iota
```
mdlayher: I think you need to add the type explicitly for these to not just be of type `int`.

libvirt.go
```diff
@@ -399,6 +413,42 @@ func (l *Libvirt) Run(dom string, cmd []byte) ([]byte, error) {
...
+// cleaning up snapshot metadata. For more information on available
+// flags, see UndefineFlag*.
+func (l *Libvirt) Undefine(dom string, flags UndefineFlags) error {
+	d, err := l.lookup(dom)
+	if err != nil {
+		return err
+	}
+
+	payload := struct {
+		Domain Domain
```
mdlayher: No padding this time? :)

benlemasurier: =p

libvirt.go
benlemasurier: oops, nice catch (it does work, but much better to be explicit)!

Issue #16, opened 2016-10-04
ci: remove scripts/golint.sh, replace with built-in golint flag


mdlayher:
`golint` has a flag that causes it to exit with non-zero status, rendering our CI build script for this purpose obsolete.

Replace with `golint -set_exit_status ./...` in CI.

This is an easy Hacktoberfest contribution, so I'll leave this unfixed for a few days.  Feel free to submit a PR to fix this and work toward that t-shirt!

mdlayher:
Closed by #17 .

micoratih:
#17

PR #17, opened 2016-10-04
Replaced scripts/golint.sh with built-in golint flag.

mkoppmann:
Fixes issue #16.

Should I delete the scripts/golint.sh?


mkoppmann:
Thanks :tada:

PR #19, opened 2016-10-27
Adds Capabilities()

benlemasurier:
This adds a new method, Capabilities(), which returns an XML definition of the hypervisor's system
capabilities, much like `virsh capabilities`.

I've only included integration tests for these two methods due to the large amount of data returned by the capabilities call.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -170,6 +170,24 @@ const (
 	DestroyFlagGraceful
 )
 
+// Capabilities returns an XML document describing the host's capabilties.
+func (l *Libvirt) Capabilities() ([]byte, error) {
+	resp, err := l.request(constants.ProcConnectGetCapabilties, constants.ProgramRemote, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	r := <-resp
+	if r.Status != StatusOK {
+		return nil, decodeError(r.Payload)
+	}
+
+	dec := xdr.NewDecoder(bytes.NewReader(r.Payload))
+	caps, _, err := dec.DecodeString()
+
+	return []byte(caps), err
+}
+
 // Connect establishes communication with the libvirt server.
 // The underlying libvirt socket connection must be previously established.
 func (l *Libvirt) Connect() error {
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -38,6 +38,7 @@ const (
 const (
 	ProcConnectOpen              = 1
 	ProcConnectClose             = 2
+	ProcConnectGetCapabilties    = 7
 	ProcDomainGetXMLDesc         = 14
 	ProcDomainLookupByName       = 23
 	ProcAuthList                 = 66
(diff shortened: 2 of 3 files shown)


libvirt.go
```diff
@@ -142,6 +143,52 @@ const (
 	UndefineFlagNVRAM
 )
 
+// Capabilities returns an XML document describing the host's capabilties.
+func (l *Libvirt) Capabilities() ([]byte, error) {
```
mdlayher: It seems like we should probably return a struct unmarshaled from the XML, instead of the XML itself.

Maybe return both for now with a TODO to add more fields?

libvirt.go
```diff
@@ -142,6 +143,52 @@ const (
...
+	}
+
+	dec := xdr.NewDecoder(bytes.NewReader(r.Payload))
+	caps, _, err := dec.DecodeString()
+
+	return []byte(caps), err
+}
+
+// CPUFlags returns a list of CPU flags supported by the host.
+func (l *Libvirt) CPUFlags() ([]string, error) {
```
mdlayher: It feels a little odd to have our own call that isn't part of the libvirt API.

Perhaps it would be beneficial to make a higher-level type (like `qemu.Domain`) and put these kinds of functions on that type, which wraps the raw libvirt type.

libvirt.go
```diff
@@ -142,6 +143,52 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := xml.Unmarshal(resp, &caps); err != nil {
+		return nil, err
+	}
+
+	var flags []string
+	for _, f := range caps.CPU[0].Features {
```
mdlayher: Should do a bounds check before accessing slice index.

libvirt.go
```diff
@@ -142,6 +143,52 @@ const (
...
+	resp, err := l.Capabilities()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := xml.Unmarshal(resp, &caps); err != nil {
+		return nil, err
+	}
+
+	var flags []string
```
mdlayher: Pre-allocate with capacity `len(caps.CPU[0].Features)`.

libvirt.go
mdlayher: Maybe `libvirt.Libvirt` becomes `libvirt.Conn` and we introduce a new `libvirt.Manager` or similar.

benlemasurier: This could easily be left to the consumer as well. Added here because _I_ need it ;) I'd rather keep this library small and simple. How about I just pull `CPUFlags()` if we're going to stick to API-level functionality?

benlemasurier:
> Can the capabilities output be reduced in any way so we can have it verified without the integration test too?

This would be easy if XDR encoding wasn't involved. On my system, `Capabilties()` returns > 700 lines of XML. Shortening the output for an integration test would require some serious effort due to the byte boundaries/padding required by XDR encoding. I did go down this path but abandoned it because we have integration tests now :)

PR #22, opened 2016-11-12
Initial support for SASL Authentication support

vinzenz:
In this PR the basics for SASL authentication are added - This implements a pure go implementation for SASL DIGEST-MD5 authentication.


mdlayher:
Very cool.  Can you write tests for this new code as well, please? Perhaps you were planning on doing so already.

vinzenz:
Yes I was planning that I just was hoping for a conceptual ack before continuing
It'd be cool if you could look over it and comment on things you rather do differently. I will follow up with cleanup and tests on it

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+		if entry == "DIGEST-MD5" {
+			return entry, nil
+		}
+	}
+	return "", fmt.Errorf("SASL: Requested mech not supported")
+}
+
+func md5Sum(a, b []byte) []byte {
+	h := md5.New()
+	h.Write(a)
```
mdlayher: Mind doing `_, _ = h.Write(a)` so the error is explicitly ignored?  Docs for `hash.Hash` mention that `Write` never returns an error.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+			},
+		},
+	}
+}
+
+func (client *client) FindMech(mechList []string) (mech string, err error) {
+	for _, entry := range mechList {
+		// Eventually this would be the place to switch to different
+		// mech variations, but for now we only support md5
+		if entry == "DIGEST-MD5" {
```
mdlayher: Seems like this and similar types should be a set of constants.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+			return entry, nil
+		}
+	}
+	return "", fmt.Errorf("SASL: Requested mech not supported")
```
mdlayher: Lowercase `sasl` at start of error message.  Maybe toss in a `%v` to show the invalid `mechList`.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+func minUint32(x, y uint32) uint32 {
+	if x < y {
+		return x
+	}
+	return y
+}
+
+func (client *client) calcSecretHash(password string) []byte {
+	h := md5.New()
+	h.Write([]byte(client.context.Username))
```
mdlayher: I think `io.WriteString` is a little more readable for all of these `h.Write([]byte(s))` statements.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+	h.Write([]byte(":"))
+	h.Write([]byte(password))
+	x := h.Sum(nil)
+	return x[:]
+}
+
+func (client *client) calcHA1(password string) []byte {
+	ha1 := md5.New()
+	tmp := client.calcSecretHash(password)
+	if client.context.Algorithm == "md5-sess" {
```
mdlayher: This should probably be a constant.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+		ha1.Write([]byte(client.context.NonceHex))
+		ha1.Write([]byte(":"))
+		ha1.Write([]byte(client.context.CNonceHex))
+		if client.context.AuthzID != "" {
+			ha1.Write([]byte(":"))
+			ha1.Write([]byte(client.context.AuthzID))
+		}
+	}
+	client.context.HA1 = ha1.Sum(nil)
+	client.context.SessionKey = fmt.Sprintf("%x", client.context.HA1)
```
mdlayher: How about `hex.EncodeToString`?  Not a big fan of `fmt.Sprintf` when there's a stdlib function available.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+	return client.context.HA1
+}
+
+func (client *client) calcHA2() string {
+	digest := md5.New()
+	digest.Write([]byte(client.context.Method))
+	digest.Write([]byte(":"))
+	digest.Write([]byte(client.context.DigestURI))
+	if client.context.QOP != "auth" {
+		digest.Write([]byte(":00000000000000000000000000000000"))
```
mdlayher: Care to leave a comment explaining this?

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+		respHash.Write([]byte(":"))
+	}
+	respHash.Write([]byte(client.calcHA2()))
+	return fmt.Sprintf("%x", respHash.Sum(nil))
+}
+
+func (client *client) requestInfo() string {
+	password := ""
+	items := []CallbackRequestItem{
+		CallbackRequestItem{CredentialPassphrase, &password},
```
mdlayher: Use a keyed struct literal.

sasl/client.go
```diff
@@ -0,0 +1,298 @@
...
+		items = append(items, CallbackRequestItem{CredentialUsername, &client.context.Username})
+	}
+	if client.Callback(items) == nil {
+		return password
+	}
+	return ""
+}
+
+func (client *client) MakeResponse() string {
+	client.context.NC++
```
mdlayher: I think a mutex should be added to this type to make it concurrency safe.

sasl/hmac.go
```diff
@@ -0,0 +1,230 @@
...
+	"crypto/des"
+	"crypto/hmac"
+	"crypto/md5"
+	"crypto/rc4"
+	"encoding/binary"
+	"fmt"
+)
+
+var (
+	errEmptyBuffer                = fmt.Errorf("ERROR: Empty buffer")
```
mdlayher: Remove `ERROR: ` from all of these strings and make the initial letters lowercase.

sasl/hmac.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+		sec.blockSize = block.BlockSize()
+		sec.DecCipher = cipher.NewCTR(block, sec.DecKey[len(sec.DecKey)-8:]).XORKeyStream
+		block, err = des.NewCipher(sec.EncKey)
+		if err != nil {
+			return err
+		}
+		sec.EncCipher = cipher.NewCTR(block, sec.EncKey[len(sec.EncKey)-8:]).XORKeyStream
+	default:
+		return fmt.Errorf("No such cipher")
```
mdlayher: Lowercase `No`, and then show the incorrect cipher type in the error message perhaps?

sasl/hmac.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (sec *securityIntegrityContext) generateKeys() error {
+	sec.EncKey = md5Sum(sec.HA1, []byte(sealingClientServer))
+	sec.DecKey = md5Sum(sec.HA1, []byte(sealingServerClient))
+	sec.Ki = md5Sum(sec.HA1, []byte(signingClientServer))
+	sec.PeerKi = md5Sum(sec.HA1, []byte(signingServerClient))
+	return nil
+}
+
+func (sec *securityIntegrityContext) getHMAC(msg []byte) []byte {
```
mdlayher: From this point on there are a lot of various offsets and lengths that aren't obvious to someone who is unfamiliar with the code.  Mind making them constants or adding comments so the numbers aren't so magical?

sasl/hmac.go
```diff
@@ -0,0 +1,230 @@
...
+	b.Write(data[0:1])
+	binary.Write(b, binary.BigEndian, sec.SeqNum)
+	return b.Bytes(), nil
+}
+
+func (sec *securityPrivacyContext) Unwrap(data []byte) (out []byte, err error) {
+	if data == nil || len(data) <= 6 {
+		return nil, errBufferTooSmall
+	}
+	sec.PeerSeqNum++
```
mdlayher: Will probably need a mutex on this type too.

sasl/hmac.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	data = data[:len(data)-6]
+	out = make([]byte, len(data))
+	sec.DecCipher(data, out)
+	hmac := out[len(out)-10:]
+	out = out[:len(hmac)]
+
+	expectedHMAC := sec.getPeerHMAC(out)
+	if bytes.Compare(expectedHMAC, hmac) != 0 {
```
mdlayher: https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare

vinzenz:
The auth is still not working properly with the secure connection method, the sasl implementation has to be done more properly before this can be properly implemented - Therefore I will drop this PR thanks for the time spent on the PR anyway - I will leave the code in my repo though, maybe someone might benefit from it at some point

mdlayher:
@vinzenz Sounds good.  If you end up getting things working, we'd love to bring this upstream!  Let us know.

PR #24, opened 2016-12-01
Adds support for dumping domain XML

benlemasurier:
This adds support for dumping a domain's XML definition, akin to `virsh dumpxml <domain>`. The returned data is quite large so I've included an integration test rather than adding a huge blob of hex to `libvirttest`.

### DIFF
--- a/libvirt_integration_test.go
+++ b/libvirt_integration_test.go
@@ -17,6 +17,7 @@
 package libvirt
 
 import (
+	"encoding/xml"
 	"net"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


libvirt_integration_test.go
```diff
@@ -40,6 +41,26 @@ func TestDisconnectIntegration(t *testing.T) {
...
+	}
+	defer l.Disconnect()
+
+	var flags DomainXMLFlags
+	data, err := l.XML("test", flags)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var v interface{}
```
mdlayher: I assume that unmarshaling will fail if invalid XML is spit out, and this is just to test that it doesn't fail?

benlemasurier: Exactly. Libvirt changes the XML definition a bit once the domain has been defined. Otherwise it would be useful to just compare against `test-domain.xml`.

PR #26, opened 2016-12-05
Add DefineXML()

nemca:
I haven't yet written tests because i cannot take a dump.
Can you help me with dump?


libvirt.go
```diff
@@ -641,6 +649,34 @@ func (l *Libvirt) XML(dom string, flags DomainXMLFlags) ([]byte, error) {
 	return []byte(s), nil
 }
 
+// DefineXML a domain, but does not start it.
```
mdlayher: Add "defines"; should be: `DefineXML defines a domain, but does not start it.`.

libvirt.go
```diff
@@ -159,6 +159,14 @@ const (
 	UndefineFlagNVRAM
 )
 
+// DomainDefineXMLFlags specifies options available when defining a domain.
+type DomainDefineXMLFlags uint32
+
+const (
+	// DefineValidate validate the XML document against schema
```
mdlayher: *validates, add a period.

benlemasurier:
@nemca Here's the result from defining a domain named `test`.

```
00 00 00 38 20 00 80 86 00 00 00 01 00 00 01 5e
00 00 00 01 00 00 00 06 00 00 00 00 00 00 00 04
74 65 73 74 af c2 ef 71 66 e0 45 a7 a5 ec d8 ba
1e a8 17 7d ff ff ff ff
```

wireshark


I strongly suggest you go through the process of building wireshark on your mac or on a VM so you can do this yourself.

nemca:
Thanks for dump.
I made a host with wireshark and now I can shoot dumps.
But i don't understand how it works.
How a can convert dump (ex: procedure number `00 00 01 5e`) to hexadecimal for tests?

benlemasurier:
Hey @nemca, I'm not sure I understand. That is a hexadecimal dump, just prefix each byte with `0x`, e.g.,:

```
0x00 0x00 0x00 0x38
0x20 0x00 0x80 0x86
0x00 0x00 0x00 0x01
0x00 0x00 0x01 0x5e
```

nemca:
About hexadecimal numbers. I think i just confused. But now I know what to do with them. :)

PR #27, opened 2017-01-06
Adds StoragePools()

benlemasurier:
This adds the ability to list storage pools.


xmudrii:
Maybe `scripts` dir. Hmmm - idea - create `examples` folder like in [`go-qemu`](https://github.com/digitalocean/go-qemu/tree/master/examples) put there example xmls and some runable code so someone who checks this out can easy create needed resources and try `go-libvirt`.
It's not really a example but I liked idea. :+1: :D

benlemasurier:
>  I think we should probably store these XML files in a .travis directory or something to avoid repo clutter though.

Makes sense to me, I've put them in the `.travis` folder.

PR #31, opened 2017-04-03
minor improvements

vtolstov:
* many improvements


vtolstov:
@benlemasurier why not break sooner to align with libvirt? Events not say anything about it, but DomainEvent says.. Also About Undefine, why Undefine and not DomainUndefine? Later i send patch to pool and it name PoolUndefine? Why not align with libvirt?

benlemasurier:
@vtolstov It has never been a goal of this project to align with the official Libvirt APIs. The existing API is largely focused on domain operations.  We're favoring short, succinct names where no duplicate functionality exists. `Events()` takes a domain, which further describes the desired action, a paradigm that has worked well.

libvirt.go
```diff
@@ -108,6 +110,75 @@ const (
 	DomainXMLFlagMigratable
 )
 
+// DomainCreateFlags specifies options when performing a domain creation.
+type DomainCreateFlags uint32
+
+const (
```
benlemasurier: use `iota` for these constants, e.g., `DomainCreateFlagPaused = 1 << iota`...

libvirt.go
```diff
@@ -108,6 +110,75 @@ const (
...
+	DomainCreateFlagStartForceBoot = 8
+
+	// DomainCreateFlagStartValidate validate the XML document against schema
+	DomainCreateFlagStartValidate = 16
+)
+
+// DomainRebootFlagValues specifies options when performing a reboot.
+type DomainRebootFlagValues uint32
+
+const (
```
benlemasurier: same as above

libvirt.go
```diff
@@ -108,6 +110,75 @@ const (
...
+	DomainRebootFlagSignal = 8
+
+	// DomainRebootFlagParavirt use paravirt guest control.
+	DomainRebootFlagParavirt = 16
+)
+
+// DomainShutdownFlagValues specifies options when performing a shutdown.
+type DomainShutdownFlagValues uint32
+
+const (
```
benlemasurier: same as above

libvirt.go
```diff
@@ -449,22 +520,17 @@ func (l *Libvirt) Events(dom string) (<-chan DomainEvent, error) {
...
 }
 
-// Migrate synchronously migrates the domain specified by dom, e.g.,
+// DomainMigrate synchronously migrates the domain specified by dom, e.g.,
 // 'prod-lb-01', to the destination hypervisor specified by dest, e.g.,
 // 'qemu+tcp://example.com/system'. The flags argument determines the
 // type of migration and how it will be performed. For more information
 // on available migration flags and their meaning, see MigrateFlag*.
-func (l *Libvirt) Migrate(dom string, dest string, flags MigrateFlags) error {
+func (l *Libvirt) DomainMigrate(d *Domain, dest string, flags MigrateFlags) error {
```
benlemasurier: leave this API  intact, `dom string...`

libvirt.go
```diff
@@ -503,15 +569,10 @@ func (l *Libvirt) Migrate(dom string, dest string, flags MigrateFlags) error {
...
+// DomainMigrateSetMaxSpeed set the maximum migration bandwidth (in MiB/s) for a
 // domain which is being migrated to another host. Specifying a negative value
 // results in an essentially unlimited value being provided to the hypervisor.
-func (l *Libvirt) MigrateSetMaxSpeed(dom string, speed int64) error {
-	d, err := l.lookup(dom)
-	if err != nil {
-		return err
-	}
-
+func (l *Libvirt) DomainMigrateSetMaxSpeed(d *Domain, speed int64) error {
```
benlemasurier: same as above

libvirt.go
```diff
@@ -778,19 +834,44 @@ func (l *Libvirt) Undefine(dom string, flags UndefineFlags) error {
...
 }
 
-// Destroy destroys the domain specified by dom, e.g., 'prod-lb-01'.
+// DomainDestroy destroys the domain.
 // The flags argument allows additional options to be specified such as
 // allowing a graceful shutdown with SIGTERM than SIGKILL.
 // For more information on available flags, see DestroyFlag*.
-func (l *Libvirt) Destroy(dom string, flags DestroyFlags) error {
-	d, err := l.lookup(dom)
+func (l *Libvirt) DomainDestroy(d *Domain, flags DestroyFlags) error {
```
benlemasurier: same as above

libvirt.go
```diff
@@ -814,14 +895,39 @@ func (l *Libvirt) Destroy(dom string, flags DestroyFlags) error {
...
 }
 
-// XML returns a domain's raw XML definition, akin to `virsh dumpxml <domain>`.
-// See DomainXMLFlag* for optional flags.
-func (l *Libvirt) XML(dom string, flags DomainXMLFlags) ([]byte, error) {
-	d, err := l.lookup(dom)
+// DomainShutdown reboot the domain.
+// The flags argument allows additional options to be specified.
+// For more information on available flags, see DomainShutdownFlagValues*.
+func (l *Libvirt) DomainShutdown(d *Domain, flags DomainShutdownFlagValues) error {
```
benlemasurier: same as above

libvirt.go
vtolstov: this is hard, because as you see some const have holes between values, and benefits of iota not very useful.

libvirt.go
vtolstov: Sorry this is not suitable for me, because in this case i have unnecessary lookups all the time and in my case i can have two domains with the same name (that does not forbidden in libvirt because of uuid). So if this is not possible with this project - i'm maintain my own fork, sorry.

libvirt.go
benlemasurier: here's an example as defined by the libvirt source:

```
/**
 * virDomainCreateFlags:
 *
 * Flags OR'ed together to provide specific behaviour when creating a
 * Domain.
 */
typedef enum {
    VIR_DOMAIN_NONE               = 0,      /* Default behavior */
    VIR_DOMAIN_START_PAUSED       = 1 << 0, /* Launch guest in paused state */
    VIR_DOMAIN_START_AUTODESTROY  = 1 << 1, /* Automatically kill guest when virConnectPtr is closed */
    VIR_DOMAIN_START_BYPASS_CACHE = 1 << 2, /* Avoid file system cache pollution */
    VIR_DOMAIN_START_FORCE_BOOT   = 1 << 3, /* Boot, discarding any managed save */
    VIR_DOMAIN_START_VALIDATE     = 1 << 4, /* Validate the XML document against schema */
} virDomainCreateFlags;
```

libvirt.go
mdlayher: There ought to be a way to accommodate both methods, but it looks like this PR has grown far beyond "minor improvements".  Would you mind filing an issue to discuss your concerns?  That way, some discussion can happen up front and a decision can be agreed upon before more code is written.

vtolstov: Ok i'm fill the issue

libvirt.go
vtolstov: thanks, fixed

libvirt_integration_test.go
```diff
@@ -122,7 +122,7 @@ func TestStoragePoolIntegration(t *testing.T) {
 	}
 
 	wantName := "test"
-	pool, err := l.StoragePool(wantName)
+	pool, err := l.StoragePoolLooupByName(wantName)
```
Alukardd: mistake? Looup instead of Lookup

vtolstov: thanks!

vtolstov:
wait for discussion on #32

PR #35, opened 2017-05-03
Add DomainCreate method

polachok:

PR #37, opened 2017-05-04
Reboot reset

polachok:


libvirt.go
```diff
@@ -260,6 +260,22 @@ const (
...
+	RebootAcpiPowerBtn RebootFlags = 1 << iota
+	// RebootGuestAgent - use guest agent
+	RebootGuestAgent
+	// RebootInitctl - use initctl
+	RebootInitctl
+	// RebootSignal - use signal
+	RebootSignal
+	// RebootParavirt - use paravirt guest control
+	RebootParavirt
+
```
benlemasurier: remove this `\n`

libvirt.go
```diff
@@ -260,6 +260,22 @@ const (
 	StoragePoolsFlagZFS
 )
 
+// RebootFlags specifies domain reboot methods
+type RebootFlags uint32
+const (
+	// RebootAcpiPowerBtn - send ACPI event
+	RebootAcpiPowerBtn RebootFlags = 1 << iota
```
benlemasurier: `\n` after each `const` definition

libvirt.go
```diff
@@ -918,6 +934,72 @@ func (l *Libvirt) Version() (string, error) {
 	return versionString, nil
 }
 
+// Reboot reboots the domain
```
benlemasurier: this comment can be improved, it should also be a full sentence. Add a note about the behavior when flags is 0.

libvirt.go
polachok: fixed

benlemasurier:
Looks great! We've got a conflict from #35, fix that up and I'll get this merged.

Issue #40, opened 2017-06-01
Obtain statistics about domains


amandahla:
Like command:
virsh domstats domainName

From [remote_protocol.x](https://libvirt.org/git/?p=libvirt.git;a=blob_plain;f=src/remote/remote_protocol.x;hb=HEAD)
 REMOTE_PROC_CONNECT_GET_ALL_DOMAIN_STATS = 344

trapgate:
This is addressed in deb7a54ff8b8a5bdeb4faa8f555e9a7394eb5948; the entry point would be ConnectGetAllDomainStats()

PR #41, opened 2017-06-05
Add ProcDomainStats procedure #WIP (#40)

amandahla:
This is a draft for issue #40 

I need help to understand how to decode the XDR. :-( 
Please, anyone?

I put the file DomainStats.pcap that I'm using on diretory 'scripts'. 

The reply is something like:
retStats -> 
	retStats[] -> 
		dom
		params[] ->
			field
			i (or ul or s...)


mdlayher:
Leaving for @benlemasurier since he's the libvirt XDR RPC expert.

benlemasurier:
@amandahla I highly recommend you build libvirt's dissector plugin for wireshark. With the plugin you'll be able to decode and view RPC communication on the wire.

amandahla:
@benlemasurier Thanks! I'm already using. The problem is that I don't know how to deal with this structure. Especially because the last field can be number, string ...This is why I asked for help/suggestions :-)

This is what I get:
libvirt

I try use interface type for last field but it didnt work with go-xdr

trapgate:
Closing this because we now have this function in the generated code: ConnectGetAllDomainStats().

PR #42, opened 2017-06-08
Add String() method to DomainMemoryStatTag type

amandahla:
Just to add a String() method. It will be useful for collect metrics.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -362,6 +362,35 @@ const (
 	DomainMemoryStatTagNr
 )
 
+func (d DomainMemoryStatTag) String() string {
+	switch d {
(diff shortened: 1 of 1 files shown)


benlemasurier:
Considering we don't do this anywhere else within go-libvirt, I think this is best left as an implementation detail for clients of go-libvirt. A snake-case'd string representation that works well for one, may not be ideal for another.

amandahla:
Hi! I keep the snake-case because I copy from here:
https://github.com/libvirt/libvirt/blob/381b9d0b0acc32c57e5f3e13b5ff1e19667cefbd/include/libvirt/libvirt-domain.h
But I can try something at client too.

trapgate:
Closing.

PR #44, opened 2017-08-01
fix a trio of compiler warnings

trapgate:
Fix three compiler warnings caused by using format strings in calls to
testing.Error() and Fatal(). Switched to Errorf() and Fatalf().


JenniGriesmann:
👍 
let's see if I have any power here....

PR #46, opened 2017-09-14
Handle TypedParamStrings

trapgate:
Add handling for TypedParamStrings, which are different from the other
TypedParam... types in that the decoded value is variable-sized.


libvirt.go
```diff
@@ -1347,14 +1411,15 @@ func (l *Libvirt) SetBlockIOTune(dom string, disk string, limits ...BlockLimit)
...
+		payload.Params = append(payload.Params, tp)
 	}
 
+	fmt.Println(payload)
 	buf, err := encode(&payload)
 	if err != nil {
 		return err
 	}
-
+	fmt.Println(buf)
```
benlemasurier: oops?

libvirt.go
```diff
@@ -1403,20 +1468,50 @@ func (l *Libvirt) GetBlockIOTune(dom string, disk string) ([]BlockLimit, error)
 		return nil, decodeError(r.Payload)
 	}
 
-	dec := xdr.NewDecoder(bytes.NewReader(r.Payload))
-	result := struct {
-		Limits     []TypedParam
-		ParamCount uint32
-	}{}
-	_, err = dec.Decode(&result)
+	fmt.Println(r.Payload)
```
benlemasurier: oops?

trapgate: Yep, fixed. Also added a comment to the bytes that add a test for strings.

libvirt.go
```diff
@@ -1347,14 +1411,15 @@ func (l *Libvirt) SetBlockIOTune(dom string, disk string, limits ...BlockLimit)
 
 	for _, limit := range limits {
 		tp := NewTypedParamULongLong(limit.Name, limit.Value)
-		payload.Params = append(payload.Params, *tp)
+		payload.Params = append(payload.Params, tp)
 	}
 
+	fmt.Println(payload)
```
benlemasurier: this one too?

trapgate: aye

libvirt.go
benlemasurier: I think you still missed this one

benlemasurier:
:shipit:

PR #53, opened 2017-12-15
add domain scheduler parameter constants

JenniGriesmann:


libvirt.go
```diff
@@ -111,6 +111,74 @@ const (
 	DomainXMLFlagMigratable
 )
 
+// scheduler parameters
```
trapgate: Can you add to this comment the name of the function(s) that these are used with?

JenniGriesmann:
updated with what I believe are all the functions and how/where the constant are used.

PR #55, opened 2018-01-02
Generate the remaining consts.

trapgate:
There were a number of hand-written consts in go-libvirt, including flag
values for various libvirt functions. Remove these and generate them
instead, so that we now have a complete set, and the naming is
consistent. I used c-for-go to do this generation, but turned off any
cgo usage by the generated code - we don't want or need to introduce a
dependency on cgo just to get constants from C headers. All code is
still generated using 'go generate ./...', which now calls a wrapper
script for added robustness.

This change also returns to using Go types for flags for most libvirt
functions, instead of plain integers.


internal/lvgen/generate.go
```diff
@@ -393,6 +399,138 @@ func procLink() {
...
+// all seem to have one of three suffixes, so that's what we look for here.
+//
+// This code uses the loader package to load the constants file generated by
+// c-for-go, which runs against libvirt's C sources. This file is generated by
+// 'go generate ./...' prior to the lvgen/ generator being run.
+func mapFlagTypes() map[string]ast.Expr {
+	pconf := loader.Config{}
+	f, err := pconf.ParseFile("../../const.gen.go", nil)
+	if err != nil {
+		fmt.Println("failed to read constants file: ", err)
```
benlemasurier: missing `return` ?

internal/lvgen/generate.go
```diff
@@ -393,6 +399,138 @@ func procLink() {
...
+	}
+	pconf.CreateFromFiles("const", f)
+	prog, err := pconf.Load()
+	if err != nil {
+		fmt.Println("failed to load package: ", err)
```
benlemasurier: missing `return`?

libvirt.yml
```diff
@@ -0,0 +1,59 @@
...
+    # We can't use environment variables here, but we don't want to process the
+    # libvirt version installed in the system folders (if any). Instead we'll
+    # rely on our caller to link the libvirt source directory to lv_source/, and
+    # run on that code. This isn't ideal, but changes to c-for-go are needed to
+    # fix it.
+    IncludePaths: [./lv_source/include]
+    SourcesPaths:
+        - libvirt/libvirt.h
+
+TRANSLATOR:
```
benlemasurier: hey, neato

internal/lvgen/generate.go
trapgate: Hmm, yes. I turned this into a panic - that error should never happen.

internal/lvgen/generate.go
trapgate: This is also a panic now.

libvirt.yml
trapgate: Yeah, the docs are hard to read, but this is a nice utility. I like the way it handles translations - I'm only using a small part of its abilities.

PR #57, opened 2018-01-05
Make libvirtError public and add Code and Domain consts

### LINKED ISSUE #56: Question regarding to Error Handling
I see that in the rpc package, there is a private libvirtError struct, which contains the Code of the error. I would want to access it, to distinguish connection problems from errors like "domain does not exist". I find it in combination with helper functions like e.g. `IsNotFound(err error) bool` pretty helpful.

Are there some ideas floating around on how proper error handling is done? Would it make sense to make LibvirtError public and add the error code constants to the project?

rmohr:
Here a possible solution for #56.

In order to allow fine-grained error handling:
 * change libvirtError to LibvirtError and implement the error interface.
 * extract Code and DomainID constants from libvirt sources

Fixes #56.

What version of libvirt are you using for code generation for this repo?

### DIFF
--- a/const.gen.go
+++ b/const.gen.go
@@ -121,307 +121,287 @@ const (
 	DomainBlockStatsFlushTotalTimes = "flush_total_times"
 	// DomainBlockStatsErrs as defined in libvirt/libvirt-domain.h:550
 	DomainBlockStatsErrs = "errs"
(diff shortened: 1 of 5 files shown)


trapgate:
See my comments in #56; I'll wait on a full review while we discuss. Also, the last code generation was against master in the libvirt repo, but we should get away from that. Running it against the most recent release makes sense to me.

PR #58, opened 2018-01-28
Move fetching response from channel to request() method

YorikSar:
Also parse error on bad status in request method.
This pattern was scattered all over the place.

This change also allows request method to manage lifetime of response channel which will come in handy in stream support implementation.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -131,41 +131,22 @@ func (l *Libvirt) connect() error {
 
 	// libvirt requires that we call auth-list prior to connecting,
 	// event when no authentication is used.
-	resp, err := l.request(constants.ProcAuthList, constants.Program, &buf)
+	_, err = l.request(constants.ProcAuthList, constants.Program, &buf)
 	if err != nil {
 		return err
 	}
 
-	r := <-resp
-	if r.Status != StatusOK {
-		return decodeError(r.Payload)
-	}
-
-	resp, err = l.request(constants.ProcConnectOpen, constants.Program, &buf)
+	_, err = l.request(constants.ProcConnectOpen, constants.Program, &buf)
 	if err != nil {
 		return err
-	}
-
-	r = <-resp
-	if r.Status != StatusOK {
-		return decodeError(r.Payload)
 	}
 
 	return nil
 }
 
 func (l *Libvirt) disconnect() error {
-	resp, err := l.request(constants.ProcConnectClose, constants.Program, nil)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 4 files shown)


YorikSar:
Travis is failing with unrelated error:
```
processing libvirt.yml ⠙[ERR] /usr/include/features.h:145:10: cannot redefine macro __GNUC_PREREQ: argument names differ
```
I guess c-for-go doesn't like gcc installed on Travis.

trapgate:
Hi Yuriy,

The go-libvirt build got broken by a recent change to c-for-go. I've put in a workaround to get things working again, and submitted a patch upstream. If you push an empty change to this branch it should rebuild successfully.

Thank you!

YorikSar:
@trapgate Thanks. I've rebased my change on master.

rpc.go
```diff
@@ -316,7 +292,7 @@ func (l *Libvirt) deregister(id uint32) {
 // request performs a libvirt RPC request.
 // The returned channel is used by the caller to receive the asynchronous
```
trapgate: This comment should be updated now that this method isn't returning a channel.

rpc.go
```diff
@@ -316,7 +292,7 @@ func (l *Libvirt) deregister(id uint32) {
...
 // call response. The channel is closed once a response has been sent.
-func (l *Libvirt) request(proc uint32, program uint32, payload *bytes.Buffer) (<-chan response, error) {
+func (l *Libvirt) request(proc uint32, program uint32, payload *bytes.Buffer) (response, error) {
```
trapgate: This should probably return (*response, error) - the response can be quite large and we don't want to copy it.

rpc.go
```diff
@@ -344,22 +320,27 @@ func (l *Libvirt) request(proc uint32, program uint32, payload *bytes.Buffer) (<
 	defer l.mu.Unlock()
 	err := binary.Write(l.w, binary.BigEndian, p)
 	if err != nil {
-		return nil, err
+		return response{}, err
```
trapgate: With a change to return a *response these can stay nil, err.

rpc.go
YorikSar: The response struct is just a couple of words big: it contains only uint32 and slice (ptr to data + 2 ints). I figured there's no real benefit in storing this in heap. Besides we were copying this data to and from channel before anyway.

YorikSar:
@trapgate Thanks for your review. I've changed comment for request() function. Waiting for your opinion on returning response by reference.

PR #62, opened 2018-02-28
 Adds DomainUpdateDeviceFlags to translation table

benlemasurier:


libvirt.gen.go
```diff
@@ -8865,8 +8865,8 @@ func (l *Libvirt) DomainEventGraphics() (err error) {
 }
 
 // DomainUpdateDeviceFlags is the go wrapper for REMOTE_PROC_DOMAIN_UPDATE_DEVICE_FLAGS.
-func (l *Libvirt) DomainUpdateDeviceFlags(Dom Domain, XML string, Flags uint32) (err error) {
-	var buf []byte
+func (l *Libvirt) DomainUpdateDeviceFlags(Dom Domain, XML string, Flags DomainDeviceModifyFlags) (err error) {
```
trapgate: Hmm. Why did the generator change this back to a bytes.Buffer?

trapgate:
Lately I've been running the generator against the latest released libvirt (which I think is 4.0.0) before committing - it looks like you're using an older version, so there are some extra diffs in the generated files.

benlemasurier:
I was wondering the same thing, must have been the merge conflict. Sources generated from libvirt-4.0.0, pulled from ftp://libvirt.org/libvirt/libvirt-4.0.0.tar.xz
```
$ grep PACKAGE_VERSION= $LIBVIRT_SOURCE/configure
PACKAGE_VERSION='4.0.0'
```


Looks like we're good now?

PR #63, opened 2018-03-01
Add streams support to libvirt RPC API

YorikSar:
This enables methods like:
* `DomainOpenConsole`
* `StorageVolUpload`
* `StorageVolDownload`
* and some others to actually work.

To generate profile for such functions, I used annotations provided in `remote_protocol.x` in libvirt sources for each function. If `@readstream` is specified, an `io.Writer` argument is added to corresponding place in argument list. `@writestream` translates to `io.Reader` in the same manner.

I switched all generated functions to use `requestStream` even if no streams are used in them because there's no profit from deciding between `request` and `requestStream` for each one of them (`request` is a thin wrapper around `requestStream`).

`requestStream` (and `request`) remain synchronous, which means it won't return until all data is sent from provided `Reader` or received to provided `Writer`. This approach seems cleaner to me than making it asynchronous, i.e. returning `Writer` instead of getting `Reader` in arguments for outgoing stream.


YorikSar:
Oh, btw, I wanted to add some integration test for some of these functions, but I didn't come up with proper scenario for it. I'll figure something out later.

internal/lvgen/generate.go
```diff
@@ -159,13 +159,45 @@ type Case struct {
...
+	ArgsStruct     string // The name of the args struct for this procedure.
+	RetStruct      string // The name of the ret struct for this procedure.
+	ReadStreamIdx  int    // The index of read stream in function argument list
+	WriteStreamIdx int    // The index of read stream in function argument list
+}
+
+// ProcMetaGenerate is a type for flags from @generate annotation
+type ProcMetaGenerate int
+
+// @generate annotation can be none, client, server or both
```
benlemasurier: This comment doesn't describe what ProcMetaGenerate* do or indicate.

internal/lvgen/generate.go
```diff
@@ -159,13 +159,45 @@ type Case struct {
...
+	ProcMetaGenerateNone ProcMetaGenerate = iota
+	ProcMetaGenerateClient
+	ProcMetaGenerateServer
+	ProcMetaGenerateBoth
+)
+
+// ProcMetaPriority is a type for values from @priority annotation
+type ProcMetaPriority int
+
+// @priority annotation can be low or high
```
benlemasurier: what does high mean? what dose low mean? this comment doesn't help future readers.

internal/lvgen/generate.go
```diff
@@ -605,6 +654,64 @@ func parseNumber(val string) (int64, error) {
...
+		if atInd == -1 {
+			// Should be only first and last line of comment
+			continue
+		}
+		spl := strings.SplitN(line[atInd+1:], ":", 2)
+		if len(spl) != 2 {
+			return nil, fmt.Errorf("invalid annotation: %s", meta)
+		}
+		spl[1] = strings.Trim(spl[1], " ")
+		if spl[0] == "generate" {
```
benlemasurier: this large if/else if block would be better written as a `switch` statement.

rpc.go
```diff
@@ -293,11 +303,106 @@ func (l *Libvirt) deregister(id uint32) {
...
+	}
+
+	resp, err := l.getResponse(c)
+	if err != nil {
+		return resp, err
+	}
+
+	var abortOutStream chan bool
+	var outStreamErr chan error
+	if outStream != nil {
```
benlemasurier: the complexity of this block is too high

rpc.go
```diff
@@ -293,11 +303,106 @@ func (l *Libvirt) deregister(id uint32) {
...
+			for {
+				select {
+				case <-abortOutStream:
+					err = l.sendPacket(serial, proc, program, nil, Stream, StatusError)
+					break
+				default:
+				}
+				n, err = outStream.Read(buf)
+				if err != nil {
+					if err == io.EOF {
```
benlemasurier: return as early as possible here
```go
outstreamErr <- l.sendPacket(serial, proc.program, nil, Stream, StatusOK)
return
```

internal/lvgen/generate.go
YorikSar: These annotations (`@generate` and `@priority`) are not used in our code, only in libvirt, really. I guess I can copy description from `remote_protocol.x` here.

internal/lvgen/generate.go
YorikSar: tbh, these types won't help future readers as well. I guess I better get rid of them.

internal/lvgen/generate.go
YorikSar: Yeah, thanks.

rpc.go
YorikSar: I wanted to avoid repeating this snippet all over the place. I'll factor out this peace into separate function.

YorikSar:
@benlemasurier Thanks for review, will address comments shortly.

YorikSar:
And there was me thinking that my approach to drain callback channel would work...

internal/lvgen/generate.go
```diff
@@ -552,22 +560,35 @@ func AddEnumVal(name, val string) error {
 	if err != nil {
 		return fmt.Errorf("invalid enum value %v = %v", name, val)
 	}
-	return addEnumVal(name, ev)
+	return addEnumVal(name, ev, nil)
+}
+
+// AddEnumValMeta will add a new enum value with attached metadata to the list.
```
trapgate: I'd like to see a comment here or somewhere else in this file describing what we mean by metadata, with an example from the libvirt sources. Just the comment block from one of the remote procedure enum values would do. Metadata is such a generic term otherwise.

internal/lvgen/lvlexer.go
```diff
@@ -259,12 +259,17 @@ func lexText(l *Lexer) stateFn {
 
 // lexBlockComment is used when we find a comment marker '/*' in the input.
 func lexBlockComment(l *Lexer) stateFn {
+	metadataComment := strings.HasPrefix(l.input[l.pos:], "/**")
 	for {
 		if strings.HasPrefix(l.input[l.pos:], "*/") {
 			// Found the end. Advance past the '*/' and discard the comment body.
```
trapgate: nit: Update this comment now that we're not just discarding the comment body.

YorikSar:
@trapgate @benlemasurier Please take another look at this PR - I've addressed all your comments.

rpc.go
```diff
@@ -293,11 +295,104 @@ func (l *Libvirt) deregister(id uint32) {
...
+		if err != nil {
+			return response{}, nil
+		}
+	}
+
+	return resp, nil
+}
+
+func (l *Libvirt) sendStream(serial uint32, proc uint32, program uint32, stream io.Reader, abort chan bool) error {
+	buf := make([]byte, 1<<22-24)
```
benlemasurier: what's going on here?

rpc.go
benlemasurier: this could still benefit from simplification

YorikSar: I'm not sure I see how this can be made simpler... Could you please elaborate?

rpc.go
YorikSar: Just making sure packet with header won't exceed 4 Mb. I've seen such limit in constants in libvirt and I'm not sure how it works, but I think it's OK to use 4 Mb as a chunk size.

rpc.go
trapgate: The main problem is too many ifs make the code harder to reason about than it could be. A couple of simple changes that would help: 

in the for loop, change the second if to:
`if resp.Status != StatusContinue { 
    break 
}
if inStream != nil ....`

Define a function pointer called `abortOutStream()` above:

`abortOutStream := func() {}`

Then inside this block where you know outStream != nil, do this:

`abortOutStream = func() {abortOutStream <- true}`

That lets you remove two if statements, since it's always safe to call that function.

You could eliminate the last if too, by defering the logic it contains from here, but that's not as clear an improvement. Up to you.

rpc.go
trapgate: That's definitely worthy of a comment. Also I don't like using bit shifting to calculate consts - it's not good for readability. Normally I'd do `4 * units.MiB`, but go-libvirt doesn't have the units package at the moment.

Finally, if the `-24` is really needed, can it be expressed as `len(something)`? If it's not really needed, it should be dropped.

vtolstov:
@YorikSar can you complete pr to get it merged?

YorikSar:
@vtolstov yes, I will address all comments on this or next week.

rpc.go
YorikSar: I'm sorry, I see now how I should've done this. I hope, new version will look simpler and easier for you too.

rpc.go
YorikSar: Yes, I should've made this line more understandable from the beginning. Fixed now.

rpc.go
```diff
@@ -293,11 +295,104 @@ func (l *Libvirt) deregister(id uint32) {
...
+				// StatusError is handled in getResponse, so this is StatusOK, end of stream
+				break
+			}
+		}
+	}
+
+	if outStream != nil {
+		err = <-outStreamErr
+		if err != nil {
+			return response{}, nil
```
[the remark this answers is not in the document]
YorikSar: Good catch, fixed.

YorikSar:
@benlemasurier @trapgate I hope I've addressed all issues now. Please take another look.

YorikSar:
Added a test, found an issue with libvirtd breaking rpc protocol, fixed it.

trapgate:
Thanks for adding that test, and for sticking with this one.

YorikSar:
@trapgate thanks for being patient with me :)
I'm glad we can finally switch back from my fork :)

study777:
@YorikSar    

Sorry to bother you. I have a problem
 DomainOpenConsole      IS       virsh   console  domainname ?
thanks

PR #66, opened 2018-03-20
Add helper function for error detection

### LINKED ISSUE #65: is there any way to extract error code?
When calling a RPC method of libvirt, the returned error contains an error code used to indicate the error types, which is helpful.

I found there's no way to extract the error code from method call, cause the error returned in this package only contains the error message, which is not friendly for error cause tracing.

Error handling in this package(https://github.com/digitalocean/go-libvirt/blob/master/rpc.go#L355):
```go
// decodeError extracts an error message from the provider buffer.
func decodeError(buf []byte) error {
    var e libvirtError

    dec := xdr.NewDecoder(bytes.NewReader(buf))
    _, err := dec.Decode(&e)
    if err != nil {
        return err
    }

    if strings.Contains(e.Message, "unknown procedure") {
        return ErrUnsupported
    }

    return errors.New(e.Message)
}
```

As the code above shows, the actual error returned was constructed by only the `e.Message`, `e.Code` was dropped.

I want to write a code to lookup domain first, then compare the error code to determine whether to define a new domain, something like this:

```go
dom, err := l.DomainLookupByName("test")
if err != nil {
    if libvirtError, ok := err.(libvirt.Error); ok {
        if libvirtError.Code == libvirt.ERR_NO_DOMAIN {
            ...
        }
    }
}
```

Is there any plan to support error code extraction?

liaoishere:
Use c-for-go to convert error codes defined in virt-error.h in libvirt source code. So we can use them for comparison.

Add `ErrIsNotFound()` and `ErrIsOK` function to detect libvirt's ERR_NO_DOMAIN and ERR_OK

resolve: #65 

Signed-off-by: Penghui Liao <liaoishere@gmail.com>


liaoishere:
With this PR, we can extract the libvirt error code for comparison.  Code example:

```golang
package main

import (
    "fmt"
    "log"
    "net"
    "time"

    "github.com/digitalocean/go-libvirt"
)

func isNotFound(err error) bool {
    libvirtError, ok := err.(*libvirt.Error)
    if ok {
        fmt.Println("error code:", libvirtError.Code)
        return libvirtError.ErrorCode() == libvirt.ErrNoDomain
    }
    return false
}

func main() {
    c, err := net.DialTimeout("unix", "/var/run/libvirt/libvirt-sock", 2*time.Second)
    if err != nil {
        log.Fatalf("failed to dial libvirt: %v", err)
    }
    l := libvirt.New(c)
    if err := l.Connect(); err != nil {
        log.Fatalf("failed to connect: %v", err)
    }

    _, err = l.DomainLookupByName("DOMIAN_NAME_NOT_EXIST")
    if err != nil {
        if isNotFound(err) {
            fmt.Println("domain not found")
        } else {
            log.Fatal(err)
        }
    }

    if err := l.Disconnect(); err != nil {
        log.Fatalf("failed to disconnect: %v", err)
    }
}
```

the function `isNotFound` above can be used to check whether the error can be ignored

rpc.go
```diff
@@ -113,6 +114,29 @@ type libvirtError struct {
...
+
+	return false
+}
+
+// ErrIsNotFound detects libvirt's ERR_NO_DOMAIN.
+func ErrIsNotFound(err error) bool {
+	return checkError(err, ErrNoDomain)
+}
+
+// ErrIsOK detects libvirt' ERR_OK.
```
heha37: libvirt's

heha37: typo: libvirt's

liaoishere:
As discussed in #56, I add a helper function in the package for error detection and keep `libvirtError` private

const.gen.go
```diff
@@ -1963,3 +1963,193 @@ const (
 	StreamEventError    StreamEventType = 4
 	StreamEventHangup   StreamEventType = 8
 )
+
+// ErrorLevel as declared in libvirt/virterror.h:44
```
trapgate: I don't want to make these error types part of go-libvirt's public API. Can you adjust the yaml file to make them private? (I think the unexport transform in c-for-go will work here.) This applies to all the types from virterror.h.

rpc.go
```diff
@@ -113,6 +114,28 @@ type libvirtError struct {
 	Level    uint32
 }
 
+func (e libvirtError) Error() string {
+	return fmt.Sprintf("%s, code: %d, domain id: %d", e.Message, e.Code, e.DomainID)
```
trapgate: I don't think that displaying the error numbers here is helpful to the end user. This should just `return e.Message.`

rpc.go
```diff
@@ -113,6 +114,28 @@ type libvirtError struct {
...
+	}
+	return false
+}
+
+// ErrIsNotFound detects libvirt's ERR_NO_DOMAIN.
+func ErrIsNotFound(err error) bool {
+	return checkError(err, ErrNoDomain)
+}
+
+// ErrIsOK detects libvirt's ERR_OK.
```
trapgate: This shouldn't exist - if libvirt returns ERR_OK, the Go wrapper functions should always return a nil error. It'd be poor go style to have to call an error classifier to determine whether there's an error at all.

rpc.go
```diff
@@ -113,6 +114,28 @@ type libvirtError struct {
...
+func checkError(err error, expectedError ErrorNumber) bool {
+	e, ok := err.(libvirtError)
+	if ok {
+		return e.Code == uint32(expectedError)
+	}
+	return false
+}
+
+// ErrIsNotFound detects libvirt's ERR_NO_DOMAIN.
+func ErrIsNotFound(err error) bool {
```
mdlayher: Nit: I'd drop the "Err" prefix, and the typical style would be "IsNotExist", as is done in `os` and the like: https://golang.org/pkg/os/#IsNotExist

But I'll leave it up to @trapgate if he'd prefer this or not.

trapgate: Good suggestion, thanks!

Issue #51, opened 2017-12-01
qemu://session?


lostinblue:
I don't want to run VMs as root, one would think naturally. When I specify the user socket I get:

````
2017/12/01 18:53:19 failed to connect: internal error: unexpected QEMU URI path '/system', try qemu:///session
````

So I searched through the code for system but I could not find anything that would be obvious that I could change to session. Any guidance on this topic would be appreciated for both this library and the qemu one you maintain. If this a repeat I apologize, but I did look through the issues of both libraries and did not see anything that was obviously relevant.

benlemasurier:
Hi @lostinblue, this may help: https://libvirt.org/uri.html#URI_qemu

What is the full connection URI you are using?

lostinblue:
I have been studying that page and have not had luck unfortunately. 

I have tried qemu for the network and session for the address but that had failure. So I tried using the socket, by doing unix network and using /run/user/1000/libvirt/libvirt-sock as the address which generated this error. I tried also libvirtd as the network which has the same failure as qemu as the network. So I'm not sure if the URI page is helpful unfortunately. Thanks for the quick response.

lostinblue:
Oh I got it working with the libvirt/libvirt-go package but I would rather use a pure go implementation. I'm guessing its somehow hardcoded to use the root system QEMU path which seems like an odd choice because of potential VM breakouts leading to full root access of the hypervisor.

benlemasurier:
Can you give me the exact connection URI you're using, even better a sample piece of code that's failing?

lostinblue:
Thanks for taking time to help me out with this issue, I would really like to use and contribute to this library. 

So I'm using the very basic example code:

````
package main

import (
	"fmt"
	"log"
	"net"
	"time"

	"github.com/digitalocean/go-libvirt"
)

func main() {
	//c, err := net.DialTimeout("tcp", "127.0.0.1:16509", 2*time.Second)
	//c, err := net.DialTimeout("tcp", "192.168.1.12:16509", 2*time.Second)
	c, err := net.DialTimeout("unix", "session", 2*time.Second)
	if err != nil {
		log.Fatalf("failed to dial libvirt: %v", err)
	}

	l := libvirt.New(c)
	if err := l.Connect(); err != nil {
		log.Fatalf("failed to connect: %v", err)
	}

	v, err := l.Version()
	if err != nil {
		log.Fatalf("failed to retrieve libvirt version: %v", err)
	}
[...]

PR #67, opened 2018-03-22
Deprecate the old pre-generation APIs

trapgate:
- Marked all the remaining old APIs from before the code-generator as
deprecated. These will be removed from the go-libvirt library as soon as
we're moved all the DO code over to use the generated APIs.
- Consolidated a couple of APIs whose implementation was split into
multiple files.
- Removed a couple of Migration-related calls we're no longer using.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -27,7 +27,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"net/url"
 	"sync"
 
 	"github.com/davecgh/go-xdr/xdr2"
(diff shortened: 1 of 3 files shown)


libvirt.go
```diff
@@ -85,11 +84,38 @@ func (l *Libvirt) Capabilities() ([]byte, error) {
...
 // The underlying libvirt socket connection must be previously established.
 func (l *Libvirt) Connect() error {
-	return l.connect()
+	payload := struct {
+		Padding [3]byte
+		Name    string
+		Flags   uint32
+	}{
+		Padding: [3]byte{0x1, 0x0, 0x0},
+		Name:    "qemu:///system",
```
benlemasurier: probably not the best time to address this, but: https://github.com/digitalocean/go-libvirt/issues/51

trapgate: Oh, yeah, that'd be a good one to address. I'll put together another PR for that in the next little while.

PR #81, opened 2019-04-05
Lifecycle events

lebauce:
This PR allows receiving lifecycle events by adding a `LifecycleEvents` method that returns a channel of events, similar to the `Events` method.

```
	events,` _ := l.LifecycleEvents()

	for event := range events {
		switch libvirt.DomainEventType(event.Event) {
		case libvirt.DomainEventStarted:
			log.Printf("Started domain %+v", event.Dom)
		case libvirt.DomainEventStopped:
			log.Printf("Stopped domain %+v", event.Dom)
		}
	}
```

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -30,12 +30,17 @@ import (
 	"sync"
 
 	"github.com/digitalocean/go-libvirt/internal/constants"
-	"github.com/digitalocean/go-libvirt/internal/go-xdr/xdr2"
+	xdr "github.com/digitalocean/go-libvirt/internal/go-xdr/xdr2"
 )
 
 // ErrEventsNotSupported is returned by Events() if event streams
 // are unsupported by either QEMU or libvirt.
 var ErrEventsNotSupported = errors.New("event monitor is not supported")
+
+// internal event
+type event interface {
+	GetCallbackID() uint32
+}
 
 // Libvirt implements libvirt's remote procedure call protocol.
 type Libvirt struct {
@@ -50,7 +55,7 @@ type Libvirt struct {
 
 	// event listeners
 	em     sync.Mutex
-	events map[uint32]chan *DomainEvent
+	events map[uint32]eventStream
 
 	// next request serial number
 	s uint32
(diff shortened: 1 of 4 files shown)


lebauce:
@trapgate Can you please have a look ? Thanks !

libvirt.go
```diff
@@ -201,19 +216,74 @@ func (l *Libvirt) Events(dom string) (<-chan DomainEvent, error) {
...
+			c <- e.(*DomainEventCallbackLifecycleMsg).Msg
 		}
 	}()
 
 	return c, nil
 }
 
+// eventCallbackRegister registers a callback for the events of type `eventID`.
+// If `domain` is passed, the callback will be called only for the events of the domain.
+func (l *Libvirt) eventCallbackRegister(eventID int64) (chan event, error) {
```
trapgate: This routine exists in the generated code; it's called `ConnectDomainEventCallbackRegisterAny`: https://github.com/digitalocean/go-libvirt/blob/07f4be593e4b39f221afb121fa7c9c22c6bb36dd/libvirt.gen.go#L13455 It returns a CallbackID, not a stream of events, but this code should use that routine instead.

rpc.go
```diff
@@ -104,6 +104,18 @@ type response struct {
 	Status  uint32
 }
 
+// event stream associated with a program and a procedure
+type stream struct {
+	events  chan event
+	proc    uint32
```
trapgate: This should be called deregisterProc, and please comment these fields too.

rpc.go
```diff
@@ -187,11 +199,22 @@ func (l *Libvirt) callback(id uint32, res response) {
 }
 
 // route sends incoming packets to their listeners.
-func (l *Libvirt) route(h *header, buf []byte) {
+func (l *Libvirt) route(h *header, buf []byte) error {
```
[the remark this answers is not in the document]
lebauce: Sorry, was an omission from a previous version. Removed

libvirt.go
lebauce: It's indeed a lot better

trapgate:
This is a nice change to have, thank you!

lebauce:
@trapgate Thank you for this lib, it's reallly useful to our project.

PR #83, opened 2019-06-21
Send remaining bytes on io.EOF

### LINKED ISSUE #82: StorageVolUpload corrupts image
I try to upload a Ubuntu image to libvirt via StorageVolUpload. If I upload the image from a file it does work but if I upload it from a http stream the image gets corrupted.
Strangely it does work from http stream as well if I use a big buffered reader in between.
Does anyone have an idea whats going on here?

Here are the steps to reproduce the behaviour:
* download the image
```
curl -O https://cloud-images.ubuntu.com/minimal/releases/disco/release-20190605/ubuntu-19.04-minimal-cloudimg-amd64.img
```

* download test code [main.txt](https://github.com/digitalocean/go-libvirt/files/3311212/main.txt)
```
curl -o main.go https://github.com/digitalocean/go-libvirt/files/3311212/main.txt
```

* run test
```
go run main.go
```

* compare downloaded files in /var/lib/libvirt/images (or where ever your default pool stores the files)
```
sha256sum /var/lib/libvirt/images/image_from_*
```

dsbrng25b:
Fix #82

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -411,7 +411,8 @@ func (l *Libvirt) sendStream(serial uint32, proc uint32, program uint32, stream 
 		default:
 		}
 		n, err := stream.Read(buf)
-		if err != nil {
+		// If there are remaining bytes in the buffer (n > 0) on io.EOF they first has to be sent. io.EOF is then handeld with the next call to Read
+		if err != nil && !(err == io.EOF && n > 0) {
 			if err == io.EOF {
 				return l.sendPacket(serial, proc, program, nil, Stream, StatusOK)
 			}
--- a/AUTHORS
+++ b/AUTHORS
@@ -19,3 +19,4 @@ Amanda Andrade <amanda.andrade@serpro.gov.br>
 Geoff Hickey <ghickey@digitalocean.com>
 Yuriy Taraday <yorik.sar@gmail.com>
 Sylvain Baubeau <sbaubeau@redhat.com>
+David Schneider <dsbrng25b@gmail.com>


rpc.go
```diff
@@ -411,7 +411,8 @@ func (l *Libvirt) sendStream(serial uint32, proc uint32, program uint32, stream
 		default:
 		}
 		n, err := stream.Read(buf)
-		if err != nil {
+		// If there are remaining bytes in the buffer (n > 0) on io.EOF they first has to be sent. io.EOF is then handeld with the next call to Read
+		if err != nil && !(err == io.EOF && n > 0) {
```
trapgate: You don't need to make this `if` so complicated - instead just change the sendPacket call in the `if err == io.EOF` clause so that it sends `buf[:n]` instead of `nil`. That will work for any value of `n` including 0, and won't require another call to this function to report the EOF. Can you also update the comment, and wrap it at 80 chars?

dsbrng25b: Yes, the if condition was really ugly and the comment as well.
First I also had that idea, but unfortunately sending the remaining data together with StatusOK does not work because sendPacket does first send the header and if we send StatusOK we already mark the end of the stream before sending the payload.
I checked how this is solved in io.Copy and applied the same logic which means that we first send bytes and then check for errors.

dsbrng25b: I force pushed the changes to my master branch but it seems that the pull request does not recognize them.
Do you have an idea how to solve this?

benlemasurier: The PR should pick up new changes. Your master branch is still only showing a single commit against master: https://github.com/dsbrng25b/go-libvirt/commits/master

dsbrng25b: Yes, there is only one commit, because I used `git commit --amend` and force pushed to my master. Normally the PR should pick up the changes, but it seems not to work in that case.
If you compare the commits of the pull request https://github.com/digitalocean/go-libvirt/pull/83/commits and the commits in my master https://github.com/dsbrng25b/go-libvirt/commits/master you see that they are not the same.
Maybe it would help to close and reopen the PR?

trapgate: That's probably the easiest way to resolve this, but in general force-pushing to a branch after you generate a PR is problematic, and even when it works it makes it harder for reviewers to see how you responded to comments. It's much simpler to just push more transactions and let git do its thing.

dsbrng25b: @trapgate My intent was to keep the commit history clean but I see that it is problematic for the reviewing process. I closed the PR. Do you have enough permission to reopen it?

trapgate: I could re-open it, but that wouldn't fix the problem - GitHub would still have the wrong set of changes. I opened a new PR, which I can merge, but if you want the PR to show up in your own name (which I completely understand), feel free to open a fresh PR of your own and I'll merge that one instead. Thank you for fixing this bug, too!

PR #84, opened 2019-06-25
Send data before handling errors

trapgate:
If Read receives some data together with io.EOF the data has to be sent before
sending StatusOK or StatusError which indicates the end of the stream.


dsbrng25b:
Fix #82
@trapgate As a question to your offer in #83: its fine if you merge this PR.

Issue #85, opened 2019-07-12
telegraf libvirt plugin crashes in rpc.go for some VMs


ahothan:
Here is the back trace:

Jul  6 09:35:00 CVIM-compute-1 telegraf: panic: runtime error: makeslice: len out of range
Jul  6 09:35:00 CVIM-compute-1 telegraf: goroutine 189 [running]:
Jul  6 09:35:00 CVIM-compute-1 telegraf: github.com/influxdata/telegraf/vendor/github.com/digitalocean/go-libvirt.(*Libvirt).listen(0xc42073c230)
Jul  6 09:35:00 CVIM-compute-1 telegraf: /builddir/build/src/github.com/influxdata/telegraf/vendor/github.com/digitalocean/go-libvirt/rpc.go:179 +0xb9
Jul  6 09:35:00 CVIM-compute-1 telegraf: created by github.com/influxdata/telegraf/vendor/github.com/digitalocean/go-libvirt.New
Jul  6 09:35:00 CVIM-compute-1 telegraf: /builddir/build/src/github.com/influxdata/telegraf/vendor/github.com/digitalocean/go-libvirt/libvirt.go:549 +0x25d
Jul  6 09:35:00 CVIM-compute-1 systemd: telegraf.service: main process exited, code=exited, status=2/INVALIDARGUMENT
Jul  6 09:35:00 CVIM-compute-1 systemd: Unit telegraf.service entered failed state.

This only happens for some VMs and not other VMs.

benlemasurier:
Fixed by PR #86

PR #86, opened 2019-07-13
Fix rpc.go crash (issue #85)

ahothan:
https://github.com/digitalocean/go-libvirt/issues/85

This patch fixes the occasional crash that happens in rpc.go.
The crash occurs when a reply is coming back fragmented and socket read returns an incomplete
length field or header field.
The original code is incorrect as the read loop writes at the start of the buffer at every iteration when it should write at the next position (passed the bytes already read).
ReadFull() will actually do just what is needed.

I have tested the patch and it fixes the crash (which could be reproduce very easily with prior code).


benlemasurier:
Thank you!

PR #88, opened 2019-07-17
Unregister streams and callbacks when disconnected

lebauce:


vtolstov:
looks good to me

lebauce:
Could an other reviewer please take a look at this ? Thanks

lebauce:
@berrange @trapgate @vrgakos Would is it be possible to review and merge this one if it's ok ? Thanks !

trapgate:
Hey @lebauce, thank you for the PR. This is one of a few bugs that strongly suggest that we need to rework the way that go-libvirt handles events. The current support is limited to a small subset of the events supported by libvirt itself, and the way that event support was added to go-libvirt has made possible some subtle bugs where a client can deadlock the go-libvirt library simply by doing things in a different order. We're in the midst of reworking the rpc code, and I'll make sure that we get the result of your change even though the implementation will be different.

benlemasurier:
resolved by #113

PR #98, opened 2020-03-16
fix deadlock by channel waitting 

WanLinghao:
This bug was due to channel waitting each other. 
Let's assume a function try to listen events on lifecycle events channel, at the same time, it also invokes   a function at following process, which looks like:
for event := range LifecycleEvents() {
   timeConsumingFunction()
    DomainGetVcpusFlags
}
It could cause deadlock, the root of deadlock is the route() function contains both event channel and callback channel process logic. If the for loop above pulls a event from event channel, and at the time of execution of  timeConsumingFunction, more event comes. Then the route() funtion will try send those new events via event channel. But nobody could recieve those events. However, at the same time when the DomainGetVcpusFlags function got to be executed, it also hang for route() function updating the callback channel. The sad thing is, route() function would never got there since it traps in waitting.
This is how the deadlock happens, and my pr could fix it by putting both event send code into goroutine.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -153,6 +153,10 @@ func IsNotFound(err error) bool {
 // listen processes incoming data and routes
 // responses to their respective callback handler.
 func (l *Libvirt) listen() {
+	evnetWaitChan := make(chan bool)
+	go func(){
+		evnetWaitChan <- true
+	}
 	for {
 		// response packet length
 		length, err := pktlen(l.r)
@@ -184,7 +188,7 @@ func (l *Libvirt) listen() {
 		}
 
 		// route response to caller
-		l.route(h, buf)
+		evnetWaitChan = l.route(h, buf, evnetWaitChan)
 	}
 }
 
@@ -206,7 +210,7 @@ func (l *Libvirt) callback(id uint32, res response) {
 }
 
 // route sends incoming packets to their listeners.
-func (l *Libvirt) route(h *header, buf []byte) {
+func (l *Libvirt) route(h *header, buf []byte, previousWaitChan chan bool) chan bool {
 	// route events to their respective listener
 	var streamEvent event
 	switch {
@@ -219,10 +223,9 @@ func (l *Libvirt) route(h *header, buf []byte) {
 	if streamEvent != nil {
 		err := eventDecoder(buf, streamEvent)
 		if err != nil { // event was malformed, drop.
-			return
+			return previousWaitChan
 		}
-		l.stream(streamEvent)
-		return
+		return 	l.stream(streamEvent, previousWaitChan)
 	}
 
 	// send responses to caller
@@ -231,6 +234,7 @@ func (l *Libvirt) route(h *header, buf []byte) {
 		Status:  h.Status,
 	}
 	l.callback(h.Serial, res)
+	return previousWaitChan
 }
 
 // serial provides atomic access to the next sequential request serial number.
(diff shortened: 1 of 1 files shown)


WanLinghao:
@trapgate @YorikSar PTAL

rpc.go
```diff
@@ -245,12 +247,14 @@ func (l *Libvirt) stream(e event) error {
...
-		// forever without a receiver. If that happens, this write will panic.
-		defer func() {
-			recover()
+		go func() {
+			// we close the channel in deregister() so that we don't block here
+			// forever without a receiver. If that happens, this write will panic.
+			defer func() {
+				recover()
+			}()
+			c.Events <- e
```
YorikSar: In your scenario where stream is opened for a long time, if we receive multiple stream packets while channel is blocked, we'll loose their ordering since order of execution for goroutines is not guaranteed in any way.
For example (found no stackoverflow) in this case:
```
	messages := make(chan string)
	go func() { messages <- "hello" }()
	go func() { messages <- "ping" }()
	msg := <-messages
	msg2 := <-messages
	fmt.Println(msg)
	fmt.Println(msg2)
```
(playground at https://play.golang.org/p/OU29KfWpgNs) always prints "ping" before "hello" for me.
I would guess that even if you add more data to TestVolumeUploadDownloadIntegration test and add some Sleep in the middle, you'll get data in messed up order with this patch.

It looks like we'd need to implement some sort of queueing mechanism here. Maybe add a small buffer to the stream channel and/or just reject packets if stream channel is not ready.

YorikSar:
@WanLinghao Great catch! You approach doesn't look right though, added comment inline. It would be great to cover such case with some test as well.

rpc.go
trapgate: This is true - since the new code spawns a goroutine for every response we get back from the client, it could deliver those responses out of-of-order.

WanLinghao:
@YorikSar @trapgate  Thanks for your review,  I will fix it

YorikSar:
Actually, the more I think about this, the more it seems to me that this should be handled on the caller side. Current approach would always block for every incoming event/packet unless caller consumes them as fast as they arrive. This is a backpressure problem: some use cases require all events to be received in order, some use cases allow them to come in any order, some would allow to drop events. Library has no way of knowing which case user requires.
In your case if you need all events to be received and acted upon in order, I'd suggest setting up a queue for them and filling it up in intermediate goroutine that would drain event stream from library.

WanLinghao:
In my case,  I just call the function like I show above, and the deadlock happens. It is not a busy one, just 3 VMs. I don't think it should be handled in caller side. The vendor code should provide safe and sound api, not like this, right?  People would be confused why just every thing stopped.  I still believe this is a bug and we should fix it. WDYT
> Actually, the more I think about this, the more it seems to me that this should be handled on the caller side. Current approach would always block for every incoming event/packet unless caller consumes them as fast as they arrive. This is a backpressure problem: some use cases require all events to be received in order, some use cases allow them to come in any order, some would allow to drop events. Library has no way of knowing which case user requires.
> In your case if you need all events to be received and acted upon in order, I'd suggest setting up a queue for them and filling it up in intermediate goroutine that would drain event stream from library.

rpc.go
WanLinghao: @YorikSar  you are right about the order problems. But it seems adding a small buffer just decrease the chance of deadlock, not solve it. Since if caller side consumes events slowly, new coming event could still consume all the buffer and trigger deadlock.  What if we add some coordinating mechanisms to gurantee the order of events?

WanLinghao:
@trapgate @YorikSar  Just updated my PR, now it fix the event order problems by some kind of lock chain.  It might be a little ugly, but I can't figure out any better way to fix it. I think we could also consider replacing event channel with a buffer one. But just like I said, the deadlock bug is just relieve the pain not cure it. Maybe we could just drop the event when the channel is full.

YorikSar:
Your current solution seems to emulate queue by creating a channel for each element. While it looks interesting, it seems to be a recipe for leaking those channels. I also don't think we should add this complexity to handling all RPC packets.
It looks like the only two problematic methods that can cause such race condition are Events and LifecycleEvents since they are the only ones that directly translate from internal Events channel to consumer's channel.
I think we should add either some parameter or some wrapper that would allow user to select behaviour of the queue returned by those calls: size of the buffer and whether or not events should be dropped. These decisions greatly depend on concrete caller use case.

benlemasurier:
I believe this has been resolved by #113

Issue #110, opened 2020-09-01
Support polkit authentication


WanzenBug:
Currently there is no way to use these bindings with a libvirtd that is configured to use the `polkit` authentication method.

To fix this issue, a simple call to `AuthPolkit()` before opening the connection should be enough

PR #111, opened 2020-09-01
support authentication via polkit

WanzenBug:
PolKit authentication is the default on (at least) RedHat based platforms.
Support only requires sending an "AuthPolkit" request before opening the
connection.

Fixes https://github.com/digitalocean/go-libvirt/issues/110


trapgate:
I don't think there's any reason not to - the conflicts need to be resolved, and I might suggest moving the auth code into a function which is called on each response.

dmacvicar:
Submitted https://github.com/digitalocean/go-libvirt/pull/138

This one can be closed.

WanzenBug:
Thanks @dmacvicar for moving this along.

PR #114, opened 2020-12-08
Refactor event handling

trapgate:
🔥🔥🔥 **This is an API-breaking change** 🔥🔥🔥

- The libvirt QEMU endpoints and constants that used to be hardcoded are now generated by processing an additional protocol file from the libvirt sources. Some public symbol names have changed as a result.
- This reworks event subscription to, hopefully, provide more stable operation. Event subscription lifecycle is now managed by a caller-provided context, automatically cleaning up on cancelation and disconnection. Client events streams are now independently pushed to client via an unbounded queue, preventing a single client from blocking other operations.
- Updated travis builds to use go 1.15
- Regenerated the code against libvirt 6.10, the most recent release.

API breakage:
- Events() is renamed SubscribeQEMUEvents and now requires a context.Context.
- LifecycleEvents() also now requires a context.Context.


rpc.go
```diff
@@ -188,139 +178,120 @@ func (l *Libvirt) listen() {
...
+}
 
-	_, err = l.request(stream.DeregisterProc, stream.Program, buf)
-	if err != nil {
-		return err
+// removeStream notifies the libvirt server to stop sending events for the
+// provided callback ID. Upon successful de-registration the callback handler
+// is destroyed. Subsequent calls to removeStream are idempotent and return
+// nil.
+// TODO: Fix this comment
```
benlemasurier, line 250: did i leave this? don't remember doing that...

internal/lvgen/generate.go
```diff
@@ -353,7 +382,7 @@ func fromSnakeToCamel(s string) string {
 // abbrevs is a list of abbreviations which should be all upper-case in a name.
 // (This is really just to keep the go linters happy and to produce names that
 // are intuitive to a go developer.)
-var abbrevs = []string{"Xml", "Io", "Uuid", "Cpu", "Id", "Ip"}
+var abbrevs = []string{"Xml", "Io", "Uuid", "Cpu", "Id", "Ip", "Qemu"}
```
benlemasurier, line 385: :heart:

benlemasurier (approved):
went through generator updates and scanned for any #113 changes, lgtm

rpc.go
trapgate: No, that was one of mine. I'll take care of it.

internal/lvgen/generate.go
trapgate: It _is_ better this way.

Issue #116, opened 2020-12-10
Allow to specify libvirt driver URI to connect to


baez90:
First of all: I really appreciate all your efforts to get a libvirt integration without C bindings!

Currently the `Connect() error` method uses the hardcoded libvirt driver URI `qemu:///system` which is kind of a showstopper for projects where I'd like to use this library.

After playing around a bit with the code I could not notice any problem with adding an overload like

`ConnectToURI(uri string) error` that allows for instance to use the mock driver `test:///default` or other libvirt drivers.

I'll open a PR and would be happy about any feedback!

Kind regards
Peter

benlemasurier:
I'm fine with the API changing, updating `Connect()` to `Connect(uri string)`. Including an exported `const`, such as `DefaultConnectURI = qemu://system` might also be a good idea.

baez90:
Fine for me, too.
I just thought keeping the original API isn't that bad and is not so much overhead.

I just opened the PR, just let me know if I should change it!

PR #117, opened 2020-12-10
Add overload for Connect() to allow other libvirt drivers

baez90:
Issue #116


.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
benlemasurier, line 1: did you mean to include this?

libvirt.go
```diff
@@ -39,6 +39,21 @@ import (
 // are unsupported by either QEMU or libvirt.
 var ErrEventsNotSupported = errors.New("event monitor is not supported")
 
+// LibvirtURI defines a type for driver URIs for libvirt
+// the defined constants are *not* exhaustive as there are also options e.g. to connect remote via SSH
+type LibvirtURI string
```
benlemasurier, line 44: I'm not sure we gain much with an additional type here. Not a blocker for me, happy for others to weigh in.

libvirt_test.go
```diff
@@ -34,6 +35,50 @@ func TestConnect(t *testing.T) {
 	}
 }
 
+func TestLibvirt_ConnectToURI(t *testing.T) {
```
benlemasurier, line 38: :heart:

.gitignore
baez90: no I wanted to exclude it to make sure I don't mess the repo up with some GoLand specific stuff

libvirt.go
baez90: Thought it makes stuff easier for beginners to find the included constants with the IDE and type based predictions and because it's a string anyway it does not block anyone to provide custom URIs but I could of course change this?

libvirt.go
```diff
@@ -39,6 +39,21 @@ import (
...
+
+const (
+	// QemuSystem connects to a QEMU system mode daemon
+	QemuSystem LibvirtURI = "qemu:///system"
```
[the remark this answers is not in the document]
baez90: done

libvirt.go
```diff
@@ -39,6 +39,21 @@ import (
 // are unsupported by either QEMU or libvirt.
 var ErrEventsNotSupported = errors.New("event monitor is not supported")
 
+// LibvirtURI defines a type for driver URIs for libvirt
+// the defined constants are *not* exhaustive as there are also options e.g. to connect remote via SSH
```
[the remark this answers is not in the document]
baez90: done

.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
```
benlemasurier, line 1: no need to include this file

libvirt.go
benlemasurier: good enough for me :+1:

.editorconfig
baez90: removed it again - just thought it would be nice to enforce the line length limit :blush:

benlemasurier:
lgtm - once travis[1] is happy I'll get this merged. Thanks, very much appreciated!

[1] - super slow at the moment. I'm in the process of switching over to GH Actions as we speak

libvirt.go
trapgate: I'm OK with using a type here, but the name stutters: it's `libvirt.LibvirtURI`. I suggest shortening the type name to just `URI`.

libvirt.go
```diff
@@ -39,6 +39,22 @@ import (
 // are unsupported by either QEMU or libvirt.
 var ErrEventsNotSupported = errors.New("event monitor is not supported")
 
+// LibvirtURI defines a type for driver URIs for libvirt
+// the defined constants are *not* exhaustive as there are also options
+// e.g. to connect remote via SSH
+type LibvirtURI string
```
trapgate, line 45: I'm OK with using a type here, but the name stutters: it's libvirt.LibvirtURI. I suggest shortening the type name to just URI.

trapgate:
:+1: Thanks for the PR, I think this is a nice improvement.

libvirt.go
baez90: what about `DriverURI` to avoid confusion?

benlemasurier: `ConnectURI`? naming is the worst.

baez90: :joy: true that - I'll change it to `ConnectURI`

baez90:
Tests should be fixed now, I just haven't noticed that there's not `libvirt.sock` in the test environment :sweat_smile: so I changed the test to use the TCP socket (had to update my local setup though which was the reason why I wanted to use the socket in the first place I think...) but I hope this can be merged as soon as the builds are finished.

Thanks for your tremendous fast feedback and input!

benlemasurier (approved):

MattIPv4:
Hey, @baez90 - Thanks a ton for this sweet PR! 😄

Would you please shoot me an email when you get a chance?
mcowley at digitalocean dot com 🎉

PR #120, opened 2021-01-12
Fix two race conditions related to Disconnect

trapgate:
* Because the Disconnect() implementation unsubscribed existing requests
before closing the connection to libvirt, it was possible for a new
request to be sent to libvirt in between the unsubscribe call and the
closing of the connection. Such a request would then be waiting for
responses on that closed connection, which is a quiet thing to do.

* Events had an unrelated problem, where the events channel never got
closed when the connection was closed, so they also ended up waiting
forever for a message that would never come.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -149,10 +149,11 @@ func (l *Libvirt) Connect() error {
 }
 
 // Disconnect shuts down communication with the libvirt server and closes the
-// underlying net.Conn. Ordering is important here. We want to make sure the
-// connection is closed before unsubscribing and deregistering the events and
-// requests, to prevent new requests from racing.
+// underlying net.Conn.
 func (l *Libvirt) Disconnect() error {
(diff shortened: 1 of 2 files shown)


sam-github (approved):

JenniGriesmann:
LGTM

JenniGriesmann (approved):

libvirt.go
```diff
@@ -149,8 +149,17 @@ func (l *Libvirt) Connect() error {
 }
 
 // Disconnect shuts down communication with the libvirt server and closes the
-// underlying net.Conn.
+// underlying net.Conn. Ordering is important here. We want to make sure the
```
benlemasurier, line 152: nit: order comment probably isn't API-level documentation and should probably go inside the fn

benlemasurier (approved):

benlemasurier (approved):

PR #123, opened 2021-01-21
Make libvirtError public so clients can use this

### LINKED ISSUE #121: Make error public, to allow more fine grained control for clients
This issue is a followup from: https://github.com/digitalocean/go-libvirt/issues/56

We are evaluating to use `go-libvirt` library in the upstream https://github.com/dmacvicar/terraform-provider-libvirt/pull/813

Right now, not having acess to specific error is problematic, since we leveraged these errors with the other `libvirt` golang 
library. see 

Is any chance to expose this errors? What is the valid rationale? we would be able to help to improve the library if needed.

I still agree also with @rmohr https://github.com/digitalocean/go-libvirt/issues/56#issuecomment-356212381

see for example:

https://github.com/dmacvicar/terraform-provider-libvirt/blob/master/libvirt/volume.go#L22

MalloZup:
Closes #121 #56

Inspired by #57 

used by https://github.com/dmacvicar/terraform-provider-libvirt/pull/813

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -111,8 +111,8 @@ type response struct {
 	Status  uint32
 }
 
-// libvirt error response
-type libvirtError struct {
+// LibvirtError response
+type LibvirtError struct {
 	Code     uint32
 	DomainID uint32
 	Padding  uint8
@@ -120,15 +120,15 @@ type libvirtError struct {
 	Level    uint32
 }
 
-func (e libvirtError) Error() string {
+func (e LibvirtError) Error() string {
 	return e.Message
 }
 
-// checkError is used to check whether an error is a libvirtError, and if it is,
+// checkError is used to check whether an error is a LibvirtError, and if it is,
 // whether its error code matches the one passed in. It will return false if
 // these conditions are not met.
 func checkError(err error, expectedError errorNumber) bool {
-	e, ok := err.(libvirtError)
+	e, ok := err.(LibvirtError)
 	if ok {
 		return e.Code == uint32(expectedError)
 	}
@@ -472,7 +472,7 @@ func encode(data interface{}) ([]byte, error) {
 
 // decodeError extracts an error message from the provider buffer.
 func decodeError(buf []byte) error {
-	var e libvirtError
+	var e LibvirtError
 
 	dec := xdr.NewDecoder(bytes.NewReader(buf))
 	_, err := dec.Decode(&e)
--- a/rpc.go
+++ b/rpc.go
@@ -111,8 +111,8 @@ type response struct {
 	Status  uint32
 }
 
-// LibvirtError response
-type LibvirtError struct {
+// LvirtError response
+type LvirtError struct {
 	Code     uint32
 	DomainID uint32
 	Padding  uint8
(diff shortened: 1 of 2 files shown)


MalloZup:
@trapgate  the `golint` is failing on a secondary issue about the naming of `LibvirtError`.

I think we can ingnore this . I think personally having be maintainer of go projects, golint on ci is somehow difficult to maintain.

Let me know how do you think, and if I can help best :sunflower:

benlemasurier:
>  golint on ci is somehow difficult to maintain

I disagree. `golint` enforces only a small set of reasonable requirements. PRs should easily be able to work within its boundaries without much impact on implementation details.

MalloZup:
@benlemasurier  thank you fro feedback! 
Since you belong to the maintainers of the project I respect your opinions and take my comment as personal subjective opinion. I do like `golint` also, but not on `PR`s ci :grin: . but this is my personal take.

But let's **focus on the problem** of why this PR is failing

GOLINT is failing with:
```
rpc.go:115:6: type name will be used as libvirt.LibvirtError by other packages, and that stutters; consider calling this Error

Found 1 lint suggestions; failing.
```

Now we can't rename the  struct `LibvirtError` to `Error` as golint suggest, afaik because this will conflicts with 

```golang
func checkError(err error, expectedError errorNumber) bool {
// here the pb:
	e, ok := err.(Error)  // if we rename from err.(LibvirtError)
	if ok {
		return e.Code == uint32(expectedError)
	}
	return false
}
```
this would conflict with 


I would rename it with `LvirtError` so that also golint is happy.

what is your take ?

vtolstov:
What the problem with Error type ?

MalloZup:
@vtolstov problem is if we want to use `Error` we need to redisgn a bit the `go-libivirt` to use packages. 

E.g

image

vtolstov:
yes, need to have redesign some internals

trapgate:
The public error type should be called `libvirt.Error`. The existing structure that currently has that name is created by the code generator. The original type name was `remote_error`; the code generator drops the `remote_` prefix so from all names so that the resulting names are closer to the libvirt documentation. But there are already some exceptions handled by the code generator - look at the `goEquivTypes` struct in `generate.go`.

We could choose to use the current `libvirt.Error`, but that structure has a bunch of fields for which I'm not convinced that clients will ever have a use. I'd rather return the smallest possible subset of those fields.

MalloZup:
@trapgate  ok thx for feedback. 

I don't understand some points of your answer, so let me summarize.

I think we do have following points:

- 1) Decide which name to use for such struct . 
-  1 a)    `remote_error` -> we don't conflicts with existing `libvirt.Error` which is already implemented in `remote_protocol`
-  1 b)   use `Error` but requires package layout change, add subpackage or related
-  1 c)   unify the `libvirt.Error` with the existing one? on other golang file ( I ddin't researched if it make sense )

when we we are done with 1) renaming, another choice.

2) Exposing the fields.

- 2a) expose all. see comment: https://github.com/digitalocean/go-libvirt/issues/56#issuecomment-356212381
- 2b) make some fields public and privat.e

3)Alternatives.,..
 

Thank you for feedback!

MalloZup:
hi all, any update on this? I would be really appreciate some help from you since the handling error would be for us one of the last step to our research of using this `go-libvirt` api  on terraform-libvirt.

tia :sun_with_face:

trapgate:
The public error type should be libvirt.Error. In order to resolve the conflict with the struct that currently has that name, you'll need to adjust the code generator by adding an exception to `goEquivTypes` so that `remote_error` gets named something else. `remoteError` is probably appropriate here. We need that type to unmarshall errors into, but we don't want it to be the public error type. The new `libvirt.Error` type should have as few elements as possible. I suspect all it needs to have added is the integer error code.

You're also going to need to make the integer error codes public, like the original PR in #57 did.

PR #127, opened 2021-03-26
internal/event/stream: fixes unsafe queue access

benlemasurier:
This fixes a problem in the event stream where we were unsafely accessing the internal event queue. Queue access is now entirely managed through `process()` via an additional `[]Event` channel used to `peek` at the queue.


trapgate (approved):

benlemasurier:
force-merging, travis-ci.org is dead. see #128

PR #131, opened 2021-04-22
fix incorrect test data

JenniGriesmann:
The length field of the payload for the Run test wasn't accurate (it was 0x74, the data provided was 0x7A).  Since the test previously wasn't sending any subsequent messages, it went unnoticed that it left unread bytes on the pipe that the next read would pick up thinking it was reading the next header when it wasn't.  In addition, the data itself needs to conform to XDR's 4 byte block size.

Adding the Connect/Disconnect calls to the mock libvirt is just a way to trigger the code to encounter the unexpected extra bytes on the pipe.


trapgate (approved):

JenniGriesmann:
Force merging because the zombie Travis CI is still lurking.

PR #132, opened 2021-04-29
fix client deadlock

JenniGriesmann:
The main thing from a practical perspective this fixes is the case the a client has sent a request to libvirt, it crashes or the connection to libvirt is lost in some other way, and it never responds.  In this case, the go-libvirt goroutine reading would notice the closed socket and return, but it would leave any clients waiting on the response blocked.

Note that although I added a couple more disconnect unit tests, neither actually simulates the above deadlock.  (Doing so would have required a rewrite of the libvirttest implementation to not use a pipe.)

See comments for everything updated, as there are a couple other small things related to unit tests as well.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -59,6 +59,9 @@ type Libvirt struct {
 	r    *bufio.Reader
 	w    *bufio.Writer
 	mu   *sync.Mutex
+	// closed after cleanup complete following the underlying connection to
+	// libvirt being disconnected.
+	socketCleanupDone chan struct{}
 
 	// method callbacks
 	cmux      sync.RWMutex
@@ -156,17 +159,11 @@ func (l *Libvirt) Disconnect() error {
 	if err != nil {
 		return err
 	}
-
 	err = l.conn.Close()
 
-	// close event streams
-	for _, ev := range l.events {
-		l.unsubscribeEvents(ev)
-	}
-
-	// Deregister all callbacks to prevent blocking on clients with
-	// outstanding requests
-	l.deregisterAll()
+	// wait for the listen goroutine to detect the lost connection and clean up
+	// to happen once it returns
+	<-l.socketCleanupDone
 
 	return err
 }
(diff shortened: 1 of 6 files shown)


libvirt.go
```diff
@@ -618,16 +615,32 @@ func getQEMUError(r response) error {
...
+		r:                 bufio.NewReader(conn),
+		w:                 bufio.NewWriter(conn),
+		mu:                &sync.Mutex{},
+		socketCleanupDone: make(chan struct{}),
+		callbacks:         make(map[int32]chan response),
+		events:            make(map[int32]*event.Stream),
 	}
 
-	go l.listen()
+	go func() {
```
benlemasurier, line 628: not a big deal, I'd prefer if this were in a teardown-centric fn on `Libvirt`, making this read more like:

```go
go l.waitForShutdown()
```

libvirt.go
```diff
@@ -59,6 +59,9 @@ type Libvirt struct {
 	r    *bufio.Reader
 	w    *bufio.Writer
 	mu   *sync.Mutex
+	// closed after cleanup complete following the underlying connection to
+	// libvirt being disconnected.
+	socketCleanupDone chan struct{}
```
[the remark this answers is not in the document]
JenniGriesmann: that name is because in some yet to be submitted code, all the socket stuff is moved to a socket object, and this channel only becomes the thing indicating it's done cleaning up a particular socket connection (of which you could have more than one on a single Libvirt object).  So I'm not sure shutdown is quite accurate.  Maybe disconnectDone ?

libvirt.go
JenniGriesmann: yeah... I definitely prefer listenAndWait()

libvirt.go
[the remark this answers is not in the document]
benlemasurier: `disconnected` works for me

libvirt.go
JenniGriesmann: ended up going with listenAndRoute

rpc.go
```diff
@@ -141,27 +141,34 @@ func IsNotFound(err error) bool {
...
+// isTemporary returns true if the error returned from a read is transiet.
+// If the error type is an OpError, check whether the net connection
+// error condition is temporary (which means we can keep using the
+// connection).
+// Errors not of the net.OpError type tend to be things like io.EOF,
+// syscall.EINVAL, or io.ErrClosedPipe (i.e. all things that
+// indicate the connection in use is no longer valid.)
+func isTemporary(err error) bool {
+	opErr, ok := err.(*net.OpError)
+	if ok {
```
benlemasurier, line 153: feel free to ignore; i think we can simplify this to

```go
if ok {
        return opErr.Temporary()
}

return false
```

libvirt.go
```diff
@@ -621,6 +621,23 @@ func getQEMUError(r response) error {
...
 }
 
+func (l *Libvirt) listenAndRoute() {
+	// only returns once it detects a non-temporary error related to the
+	// underlying connection
+	l.listen()
+
+	// close event streams
+	for _, ev := range l.events {
+		l.unsubscribeEvents(ev)
```
benlemasurier, line 631: not your doing, let's ignore for now - iirc, unsubscribeEvents() can hang forever.

benlemasurier (approved):
:trophy: :shipit:

rpc.go
JenniGriesmann: eh, it's fine, I have to fix the typo anyway

libvirt.go
JenniGriesmann: I shall keep that in mind for future

JenniGriesmann:
since we're missing automatic integration tests still, I did some quick and dirty manual test app verification, including verifying the deadlock was fixed.

PR #133, opened 2021-05-03
fix race in disconnect of event stream cleanup

JenniGriesmann:
I'm not sure whether this fixes the deadlock in unsubscribeEvents you mentioned previously @benlemasurier , but it does fix two things:
- the unsafe access of the events map without holding the mutex
- it doesn't attempt to send the message to libvirt to unsubscribe (because at the point this code is being hit, it means we detected the connection to libvirt is already gone)

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -273,6 +273,19 @@ func (l *Libvirt) removeStream(id int32) error {
 	return nil
 }
 
+// removeAllStreams deletes all event streams.  This is meant to be used to
+// clean up only once the underlying connection to libvirt is disconnected and
+// thus does not attempt to notify libvirt to stop sending events.
+func (l *Libvirt) removeAllStreams() {
+	l.emux.Lock()
+	defer l.emux.Unlock()
+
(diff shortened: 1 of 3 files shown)


rpc.go
```diff
@@ -273,6 +273,19 @@ func (l *Libvirt) removeStream(id int32) error {
...
+// removeAllStreams deletes all event streams.  This is meant to be used to
+// clean up only once the underlying connection to libvirt is disconnected and
+// thus does not attempt to notify libvirt to stop sending events.
+func (l *Libvirt) removeAllStreams() {
+	l.emux.Lock()
+	defer l.emux.Unlock()
+
+	for _, ev := range l.events {
+		delete(l.events, ev.CallbackID)
+		ev.Shutdown()
```
benlemasurier, line 285: I think by chance we're fine here, though it may be more future-proof to `delete` after `ev.Shutdown()`

benlemasurier (approved):
one note but I'm fine either way. lgtm

PR #134, opened 2021-05-03
add socket type to handle socket connection management

JenniGriesmann:
This change is mostly a reorganizing of code in preparation to break the external API such that we can take a Dialer as input and allow clients to use a single Libvirt object to potentially connect/disconnect/reconnect multiple times.

For now, this just tries to abstract the details around the packet level reading/writing over the socket connection to libvirt into its own type.

NOTE:  I'll add a comment to show what I mean, but there's some exported constants that I moved from the top level libvirt package to the socket package.  I don't know why they are exported or if there's a risk that any clients might be referencing them.

### DIFF
--- a/rpc.go
+++ b/rpc.go
@@ -16,95 +16,21 @@ package libvirt
 
 import (
 	"bytes"
-	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
-	"net"
 	"reflect"
 	"strings"
 	"sync/atomic"
-	"unsafe"
 
 	"github.com/digitalocean/go-libvirt/internal/constants"
 	"github.com/digitalocean/go-libvirt/internal/event"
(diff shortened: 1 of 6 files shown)


rpc.go
```diff
@@ -16,96 +16,22 @@ package libvirt
...
 	"github.com/digitalocean/go-libvirt/internal/event"
 	xdr "github.com/digitalocean/go-libvirt/internal/go-xdr/xdr2"
+	"github.com/digitalocean/go-libvirt/socket"
 )
 
 // ErrUnsupported is returned if a procedure is not supported by libvirt
 var ErrUnsupported = errors.New("unsupported procedure requested")
 
-// request and response types
-const (
```
JenniGriesmann, line 38: will moving these currently exported constants potentially impact any clients?  I would expect anyone, other than the code directly interacting with libvirt to need them, but they must be exported for some purpose.... possibly just for the unit tests?

PR #137, opened 2021-05-20
fix some stream problems

trapgate:
The queue we use for streams had a couple of problems, and was overly
complex. Simplify the queue processor, and add unit tests for this file.

### DIFF
--- a/internal/event/stream.go
+++ b/internal/event/stream.go
@@ -14,7 +14,13 @@
 
 package event
 
-import "context"
+import (
+	"context"
+)
+
+type emptyEvent struct{}
+
+func (emptyEvent) GetCallbackID() int32 { return 0 }
 
 // Stream is an unbounded buffered event channel. The implementation
 // consists of a pair of unbuffered channels and a goroutine to manage them.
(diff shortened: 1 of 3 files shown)


internal/event/stream.go
```diff
@@ -66,88 +80,63 @@ func (s *Stream) start() context.CancelFunc {
...
 	for {
-		// informs send() to stop trying
-		nctx, next := context.WithCancel(ctx)
-		defer next()
+		// This loop relies on the fact that a send to a nil channel will block
+		// forever. If we have no entries in the queue, we set the send channel
+		// to nil, so the clause that attempts to send an event to the client
+		// will never complete. Clients will never receive an emptyEvent.
+		sendCh := chan Event(nil)
+		next := Event(emptyEvent{})
```
benlemasurier, line 98: i now see how this  is clever, but I couldn't understand the purpose until seeing the simplified `<s.send(...)` case below. could just be me. `empty` might be a better name? I was confused, thinking we were somehow signaling to the client that the queue was empty upon receipt of an `emptyEvent` (which i realize makes no sense).

benlemasurier (approved):
very nice. one minor nit you can safely ignore, up to you.

internal/event/stream.go
trapgate: Ugh. Clever is a 4-letter word. I'm restructuring this slightly and updating the comments to hopefully make it clearer what's going on here.

PR #138, opened 2021-06-13
support authentication via polkit

dmacvicar:
Supersedes https://github.com/digitalocean/go-libvirt/pull/111

With the following changes:

* Rebased
* Moved authentication to its own method, as requested in https://github.com/digitalocean/go-libvirt/pull/111

Fixes https://github.com/digitalocean/go-libvirt/issues/110
Required for https://github.com/dmacvicar/terraform-provider-libvirt/pull/813


libvirt.go
```diff
@@ -113,6 +113,30 @@ func (l *Libvirt) Capabilities() ([]byte, error) {
...
+		switch auth {
+		case constants.AuthNone:
+			break loop
+		case constants.AuthPolkit:
+			_, err := l.AuthPolkit()
+			if err != nil {
+				return err
+			}
+			break loop
+		}
```
JenniGriesmann, line 135: nit: this might read slightly better to just add a default case that does a "continue", and then outside the switch just do a break.  Then you don't need multiple break statements or a label.  Just preference though, not something I feel particularly strongly about.
example:  https://play.golang.org/p/GzPizpqHuiV

libvirt.go
```diff
@@ -113,6 +113,30 @@ func (l *Libvirt) Capabilities() ([]byte, error) {
 	return []byte(caps), err
 }
 
+// called at connection time
+// it calls auth-list and authenticates if such auth type is supported
+func (l *Libvirt) authenticate() error {
+	resp, err := l.AuthList()
```
JenniGriesmann, line 119: nit: might be nice to keep the comment here about the AuthList being required even for AuthNone simply to protect any future code refactors from stumbling upon that requirement. 

> // libvirt requires that we call auth-list prior to connecting,
> // even when no authentication is used.

JenniGriesmann (approved):
Some nits, but nothing I'd block on.  So I will leave it up to you.  👍

libvirt.go
dmacvicar: Agreed it is better to read.

JenniGriesmann (approved):

dmacvicar:
Thanks! :fireworks:

PR #140, opened 2021-07-01
take dialer as input instead of already existing connection

JenniGriesmann:
This is a step in the direction of allowing go-libvirt to manage the underlying physical connection instead of tying the lifetime of the Libvirt object to the lifetime of the passed in connection.

For now, all this does is allow a client to call NewWithDialer, and deprecates the existing New function.  Using NewWithDialer allows the Libvirt Connect/Disconnect functions to be called multiple times.  It also adds two basic dialer implementations for local socket and tcp.

The New function was marked as deprecated, but for now uses a basically no-op dialer that just uses the passed in connection.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -24,7 +24,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"net"
 	"sync"
 	"time"
 
@@ -138,9 +137,7 @@ func (l *Libvirt) authenticate() error {
 	return nil
 }
 
-// ConnectToURI establishes communication with the specified libvirt driver
-// The underlying libvirt socket connection must be previously established.
-func (l *Libvirt) ConnectToURI(uri ConnectURI) error {
+func (l *Libvirt) initLibvirtComms(uri ConnectURI) error {
 	payload := struct {
 		Padding [3]byte
 		Name    string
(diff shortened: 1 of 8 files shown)


libvirt.go
```diff
@@ -138,9 +139,7 @@ func (l *Libvirt) authenticate() error {
 	return nil
 }
 
-// ConnectToURI establishes communication with the specified libvirt driver
-// The underlying libvirt socket connection must be previously established.
-func (l *Libvirt) ConnectToURI(uri ConnectURI) error {
+func (l *Libvirt) initLibvirtComms(uri ConnectURI) error {
```
trapgate, line 142: This is still using the manual payload construction from before the code generator. It should really call the generated ProcConnectOpen instead.

libvirt.go
```diff
@@ -659,22 +682,35 @@ func (l *Libvirt) waitAndDisconnect() {
...
 		callbacks:    make(map[int32]chan response),
 		events:       make(map[int32]*event.Stream),
 	}
 
-	// this starts the listening and routing
-	// TODO:  The connection and listen goroutine should be moved to the
-	//  Connect function, but this is going to impact the exported API, so
-	//  trying to do as much as possible before doing that.
-	l.socket = socket.New(conn, l)
+	//
```
trapgate, line 697: missing comment?

trapgate (approved):

libvirt.go
JenniGriesmann: I'd rather not put that in the same PR as a bunch of code re-org, but I might be willing to take bribes to put it into a follow-up PR.

libvirt.go
JenniGriesmann: ... I wrote it with invisible ink

PR #141, opened 2021-07-12
integration tests

JenniGriesmann:
re-introduce automatic integration tests using github actions.

TODOs:  
- Need to add a setup to test libvirt 6.7.0 (or some other more modern version).  The building of libvirt changed to start using meson, so adding support will require different build steps that the older versions.
- More integration tests should be added to cover the use of NewWithDialer instead of just the deprecated New function.


.github/workflows/main.yml
```diff
@@ -11,16 +11,69 @@ on:
...
+          make
+          sudo make install
+          ccache --show-stats
+          popd
+          sudo libvirtd -d -l -f $GITHUB_WORKSPACE/.github/testdata/libvirtd.conf
+          sudo virtlogd -d || true
+
+      - name: Setup test artifacts
+        env:
+          TESTDATA: ${{ github.workspace }}/.github/testdata
```
sam-github, line 68: Its not common to put test artifacts in `.github/`, and testdata is independent of github or travis (because it used to be in .travis), maybe it should just be at the top-level, where its easily viewable, maintainalbe, etc? Which would also remove the need for `TESTDATA` as an env var.

.github/workflows/main.yml
```diff
@@ -11,16 +11,69 @@ on:
...
+        run: |
+          go get -d ./...
+          sudo qemu-img create -f raw -o size=10M /var/lib/libvirt/images/test.raw
```
sam-github, line 71: Why not put this in `Makefile`, so its easy for devs to find it and run tests locally before GH actions run: `sudo make test` or `sudo make ci-test`?

JenniGriesmann: From what I've read (i.e. a very quick google), it's not possible to run these tests locally without installing another tool that helps allow you to do so.  If someone wants to check their branch passes before opening the PR, they must push their branch, and then use the github GUI to manually trigger a run.  I can add something to the README to that effect.

.github/workflows/main.yml
JenniGriesmann: So looking through all that's in there (as I didn't change any of it, just moved it from a travis directory to a github directory), the libvirtd.conf is specific to the CI test setup and run, so I think that file at least should remain alongside the workflows that use them, and not the integrations tests.

The other things I think are tied to the integration tests themselves, and I'm okay moving them closer to those; however, I don't like just dumping everything in the top level directory and cluttering it.  I'd rather put a 'testdata' directory in the top level and not the files themselves.

sam-github: > put a 'testdata' directory in the top level

Yes, that's what I'd suggest.

.github/workflows/main.yml
JenniGriesmann: It would be a nice enhancement to have things organized in a way where you can manually run the tests using a container locally; however, I don't have the bandwidth at the moment to work on something like that and would rather spend that time trying to get these things updated to be able to test more modern versions of libvirt.  I haven't found the workflow of pushing the branch and clicking the button all the burdensome or particularly slow (probably because the tests themselves take a bit simply because they compiles libvirt from source).

sam-github: ^--- I lack write access, so can't resolve this thread :-).

trapgate (approved):
:1st_place_medal:

PR #142, opened 2021-07-21
ci updates

JenniGriesmann:
- fix the fact that a linting errors were not failing ci (and the couple linting errors that were present)
- update ci to simply test the default installed versions of libvirt that come with the two versions of linux (ubuntu) runners available via github actions
- update integration tests to use the NewWithDialer function, as well as add tests specific to dialer (connect/disconnect) functionality
- fix some reconnection bugs detected by new integration tests


libvirt.go
```diff
@@ -205,7 +206,10 @@ func (l *Libvirt) Disconnect() error {
 	// before unsubscribing and deregistering the events and requests, to
 	// prevent new requests from racing.
 	_, err := l.request(constants.ProcConnectClose, constants.Program, nil)
-	if err != nil {
+
+	// syscall.EINVAL is returned when attempting to write to a lost connection,
+	// In this case we don't want to fail.
+	if err != nil && err != syscall.EINVAL {
```
sam-github, line 212: This would normally be EPIPE if an fd was connected, and then the connection was closed. EINVAL I'd expect if the fd was not initialized properly, so it was either a bad fd, or it was perhaps go's default (`0`, stdin), and then some syscall was made on stdin that is valid for a socket, but not whatever stdin is connected to in CI. So, this  might be a sign of something not right in the underlying code. I'm not objecting, maybe this is right for libvirt, but its unusual. I'd suggest confirming that .Connect() does *not* return before the socket descriptor is initialized completely, because I see a test below that does .Connect(); .Disconnect(), is it possible the Disconnect() is happening before the Connect has actually done its setup?

sam-github (approved):
The EINVAL thing makes me nervous, but the CI simplification is great.

libvirt.go
JenniGriesmann: I'll have to look it up, but back when I was working on the socket code, I dug around into how golang handles writes to already closed net.Conn.  Connect definitely does not return before the connection is fully established (because the Llibvirt.Connect function actually also sends some packets on the connection in order to establish the software-level libvirt connection as well before returning.)

JenniGriesmann: Ah, I remember this now.  There is a goroutine in the socket implementation that's always reading from the connection.  When it detects a "non-temporary error" implying a lost connection (see the isTemporary function in the socket package), it closes a channel that indicates a closed connection.  When other code later attempts to send a package on the socket type, it checks for the connection already being closed and returns that syscall.EINVAL.

sam-github: That makes sense, I'm comforted :-). Maybe worth a comment update?

JenniGriesmann: done

PR #145, opened 2022-04-05
update libvirt configuration step for new releases

connorkuehl:


benlemasurier (commented):
should `LIBVIRT_BUILD` point to `LIBVIRT_SOURCE/build` on 6.7+?

connorkuehl:
> should `LIBVIRT_BUILD` point to `LIBVIRT_SOURCE/build` on 6.7+?

No, I have follow-up patches that allowed me to complete the `go generate` workflow, and after seeing your comment it made me realize that it would be better if I just rolled them up into one PR instead of trying to get them merged separately, otherwise the README might be misleading if this merges as is.

Issue #87, opened 2019-07-14
NetworkUpdate: Operation not supported: can't update 'ip' section of network 'default'


a2geek:
I'm not entirely certain if this is a bug in my code or something is up with the NetworkUpdate call from this library. 

I'm writing some code that will ultimately be installed into a VM in order to manage multiple VM's. The base OS layer does not have any libvirt dependencies, and this is the best solution for my needs. (So thanks for that!)

When I do a NetworkUpdate to add an IP>DHCP>Host entry, an error is generated: `Operation not supported: can't update 'ip' section of network 'default'`.

What I've done:
* verified the project is on the latest code.
* extracted the specific code, and it fails.
* used TLS connections (thinking there may be a different socket file to utilize)
* implemented with the `libvirt/libvirt-go` library, and it works.
* verified this also works from the `virsh` command-line.  

Therefore, I think I've got a valid setup but haven't been able to identify what may be wrong (in either code base).

Running on Ubuntu 18.04.2.

```
$ virsh version
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
Using API: QEMU 4.0.0
Running hypervisor: QEMU 2.11.1
```
Using this library (and it fails):
```
const networkDhcpXML = "<host name='vm-8029d8c8-ee6c-45b9-6476-3ba7381c3cb5' ip='192.168.123.7'/>"
const networkName = "default"
const socket = "/var/run/libvirt/libvirt-sock"

func main() {
	conn, err := net.DialTimeout("unix", socket, 2*time.Second)
	defer conn.Close()
	if err != nil {
		panic(err)
	}
	fmt.Printf("Connected to '%s'\n", socket)

	client := libvirt.New(conn)
	err = client.Connect()
	if err != nil {
		panic(err)
	}
	fmt.Printf("Connected to libvirt\n")

	net, err := client.NetworkLookupByName(networkName)
	if err != nil {
		panic(err)
	}
	fmt.Printf("net = %v\n", net)

	xmlDesc, err := client.NetworkGetXMLDesc(net, 0)
	if err != nil {
		panic(err)
	}
	fmt.Printf("networkXMLDesc=%s\n", xmlDesc)

	cmd := uint32(libvirt.NetworkUpdateCommandAddLast)
	section := uint32(libvirt.NetworkSectionIPDhcpHost)
	//netflags := libvirt.NetworkUpdateAffectLive | libvirt.NetworkUpdateAffectConfig
[...]

PR #148, opened 2022-05-12
Add NetworkUpdateCompat

cfergeau:
NetworkUpdateCompat is a wrapper over NetworkUpdate which swaps `Command`
and `Section` when needed.
This function must be used instead of NetworkUpdate to be sure that the
NetworkUpdate call works both with older and newer libvirtd connections.

libvirt on-wire protocol had a bug for a long time where Command and Section
were reversed. It's been fixed in newer libvirt versions, and backported to
some older versions. This helper detects what argument order libvirtd expects
and makes the correct NetworkUpdate call.

Ideally 'NetworkUpdate' in internal/constants/remote_protocol.gen.go
could be renamed to 'networkUpdate' or such, and the new helper would be
'NetworkUpdate', but I don't know if it's possible to do this with the
code generator.

See https://listman.redhat.com/archives/libvir-list/2021-March/msg00054.html
https://gitlab.com/libvirt/libvirt/-/commit/94741bc53e9b9a405fae8f2e96b780c2891b2ff0
https://gitlab.com/libvirt/libvirt/-/commit/b0f78d626a18bcecae3a4d165540ab88bfbfc9ee

This fixes https://github.com/digitalocean/go-libvirt/issues/87


libvirt.go
```diff
@@ -717,3 +717,27 @@ func NewWithDialer(dialer socket.Dialer) *Libvirt {
...
+// libvirt on-wire protocol had a bug for a long time where Command and Section
+// were reversed. It's been fixed in newer libvirt versions, and backported to
+// some older versions. This helper detects what argument order libvirtd expects
+// and makes the correct NetworkUpdate call.
+func (l *Libvirt) NetworkUpdateFixed(Net Network, Command NetworkUpdateCommand, Section NetworkUpdateSection, ParentIndex int32, XML string, Flags NetworkUpdateFlags) (err error) {
+	// This is defined in libvirt/src/libvirt_internal.h and thus not available in go-libvirt autogenerated code
+	const virDrvFeatureNetworkUpdateHasCorrectOrder = 16
+	hasCorrectOrder, err := l.ConnectSupportsFeature(virDrvFeatureNetworkUpdateHasCorrectOrder)
+	if err != nil {
+		// TODO: ignore error and force hasCorrectOrder to 0?
```
connorkuehl, line 734: Looking at Libvirt.ConnectSupportsFeature, I'd vote we just propagate the error down like it is now; otherwise, with those classes of errors ((de/en)coding errors, transport layer errors), I'm guessing the subsequent calls to Libvirt.NetworkUpdate will fail for similar reasons.

libvirt.go
```diff
@@ -717,3 +717,27 @@ func NewWithDialer(dialer socket.Dialer) *Libvirt {
...
+
+// NetworkUpdateFixed is a wrapper over NetworkUpdate which swaps `Command` and `Section` when needed.
+// This function must be used instead of NetworkUpdate to be sure that the
+// NetworkUpdate call works both with older and newer libvirtd connections.
+//
+// libvirt on-wire protocol had a bug for a long time where Command and Section
+// were reversed. It's been fixed in newer libvirt versions, and backported to
+// some older versions. This helper detects what argument order libvirtd expects
+// and makes the correct NetworkUpdate call.
+func (l *Libvirt) NetworkUpdateFixed(Net Network, Command NetworkUpdateCommand, Section NetworkUpdateSection, ParentIndex int32, XML string, Flags NetworkUpdateFlags) (err error) {
```
connorkuehl, line 729: Just a suggestion, I don't think the current name is bad, but what do you think of "NetworkUpdateAdaptive" or "NetworkUpdateCompat" or something like that?

libvirt.go
```diff
@@ -717,3 +717,27 @@ func NewWithDialer(dialer socket.Dialer) *Libvirt {
...
+	// This is defined in libvirt/src/libvirt_internal.h and thus not available in go-libvirt autogenerated code
+	const virDrvFeatureNetworkUpdateHasCorrectOrder = 16
+	hasCorrectOrder, err := l.ConnectSupportsFeature(virDrvFeatureNetworkUpdateHasCorrectOrder)
+	if err != nil {
+		// TODO: ignore error and force hasCorrectOrder to 0?
+		return fmt.Errorf("failed to test if VIR_DRV_FEATURE_NETWORK_UPDATE_HAS_CORRECT_ORDER is supported:  %w", err)
```
connorkuehl, line 735:
```suggestion
		return fmt.Errorf("failed to confirm argument order for NetworkUpdate:  %w", err)
```

The C symbol is interesting at the level of abstraction that a go-libvirt developer would work at, but I'm not sure clients of the package will care too much exactly what the `#define` is called.

connorkuehl (commented):
:slightly_smiling_face:  Thanks for the patch! I just had a couple of comments.

libvirt.go
cfergeau: `NetworkUpdateCompat` is much better name, thanks!

libvirt.go
cfergeau: The reason  for the comment is that I was not sure what would happen if the remote end has no implementation for `ConnectSupportsFeature`, or if the connection does not support the feature.
For the former, I now see the method has been there forever (at least libvirt 0.3.0), so this is not going to be a real concern. And thinking more of it, I don't see why an error would be returned instead of `hasCorrectOrder = 0` when the method is present but the feature not supported. So yeah, if there's an error, it's going to be a 'bad' one, and we should return it.

libvirt.go
cfergeau: Ok, I've changed it, thanks!

libvirt.go
```diff
@@ -717,3 +717,26 @@ func NewWithDialer(dialer socket.Dialer) *Libvirt {
...
+// libvirt on-wire protocol had a bug for a long time where Command and Section
+// were reversed. It's been fixed in newer libvirt versions, and backported to
+// some older versions. This helper detects what argument order libvirtd expects
+// and makes the correct NetworkUpdate call.
+func (l *Libvirt) NetworkUpdateCompat(Net Network, Command NetworkUpdateCommand, Section NetworkUpdateSection, ParentIndex int32, XML string, Flags NetworkUpdateFlags) (err error) {
+	// This is defined in libvirt/src/libvirt_internal.h and thus not available in go-libvirt autogenerated code
+	const virDrvFeatureNetworkUpdateHasCorrectOrder = 16
+	hasCorrectOrder, err := l.ConnectSupportsFeature(virDrvFeatureNetworkUpdateHasCorrectOrder)
+	if err != nil {
+		return fmt.Errorf("failed to confirm argument order for NetworkUpdate:  %w", err)
```
sam-github, line 734:
```suggestion
		return fmt.Errorf("failed to confirm argument order for NetworkUpdate: %w", err)
```
This is really trivial, but I think there is an extra SPC character in the fmt string.

MattIPv4:
Hey, @cfergeau - Thanks a ton for this sweet PR! 😄

Would you please shoot me an email when you get a chance?
mcowley at digitalocean dot com 🎉

PR #150, opened 2022-06-22
expose Libvirt.Disconnected method

### LINKED ISSUE #149: Mentioned method Disconnected() does not exist
The documentation for the Connect() method says:

```
// Connect establishes communication with the libvirt server.
// The underlying libvirt socket connection will be created via the dialer.
// Since the connection can be lost, the Disconnected function can be used
// to monitor for a lost connection.
```

There is however no such method on libvirt.Libvirt. There is a method on socket.Socket, but the socket used is not exposed by libvirt.Libvirt.

connorkuehl:
Fixes #149 

This allows client code to monitor and react when the underlying
libvirt connection transitions into a disconnected state.

### DIFF
--- a/libvirt_test.go
+++ b/libvirt_test.go
@@ -154,6 +154,40 @@ func TestLostConnectionCleanup(t *testing.T) {
 	}
 }
 
+func TestMonitorDisconnect(t *testing.T) {
+	dialer := libvirttest.New()
+	lv := NewWithDialer(dialer)
+
+	// Haven't connected yet.
+	select {
+	case <-lv.Disconnected():
+	default:
+		t.Fatalf("chan is open but should be closed")
+	}
+
+	err := lv.Connect()
+	if err != nil {
+		t.Fatalf("connect failed: %v", err)
+	}
+
+	select {
+	case <-lv.Disconnected():
+		t.Fatalf("chan should be open but is closed")
+	default:
+	}
+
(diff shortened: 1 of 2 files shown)


libvirt_test.go
```diff
@@ -154,6 +154,40 @@ func TestLostConnectionCleanup(t *testing.T) {
...
+		t.Fatalf("chan should be open but is closed")
+	default:
+	}
+
+	err = lv.Disconnect()
+	if err != nil {
+		t.Fatalf("disconnect failed: %v", err)
+	}
+
+	select {
```
sam-github, line 184: I worry this case may be racy, is the close of the channel sync in Disconnect()? I think that there is a goro that reacts to socket close and closes the channel, so that it is not guaranteed that the chan is closed _before Disconnect() returns_. It might be better to not have the default in this select, in which case, on failure (the chan never gets closed), the test would eventually timeout and fail, but on success it will be as quick.

sam-github (approved):

libvirt_test.go
connorkuehl: Disconnect synchronizes with the goro you mentioned (Libvirt.waitAndDisconnect). The goro does some cleanup when it successfully reads from the Libvirt.socket.Disconnected chan and then it closes Libvirt.disconnected which is what Disconnect is synchronizing on before returning.

PR #162, opened 2022-08-03
integration: fix failure when there is >1 pool

connorkuehl:
This will spuriously fail on a developer's machine unless they purge
their pre-existing libvirt configuration. Instead of assuming the test
is starting from a blank slate, just assert that the thing that the test
expects to be there is there, and nothing else.


trapgate (approved):
It'd be kinder to make sure the pool created by the test gets cleaned up after every run though.

PR #173, opened 2022-09-28
Fix typo in `.github/workflows` and turn into a link

mamercad:
Fix typo is `.github/workflows` and turn it into a link.

PR #174, opened 2022-10-03
workflows typo fix

gidoBOSSftw5731:


connorkuehl:
Hi @gidoBOSSftw5731, thanks for the patch! This is already being addressed in #173.

gidoBOSSftw5731:
Ah I see, I'll close this then.

PR #175, opened 2022-10-20
export a convenient Libvirt.IsConnected method

connorkuehl:
This reduces the amount of boilerplate required for client code to reference the connection state during error handling code paths, especially if the client code hasn't built a reconnect mechanism around the underlying Libvirt.Disconnected chan.

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -241,6 +241,17 @@ func (l *Libvirt) Disconnected() <-chan struct{} {
 	return l.disconnected
 }
 
+// IsConnected indicates whether or not there is currently a connection to
+// libvirtd.
+func (l *Libvirt) IsConnected() bool {
+	select {
+	case <-l.Disconnected():
+		return false
+	default:
+		return true
+	}
+}
+
 // Domains returns a list of all domains managed by libvirt.
 //
 // Deprecated: use ConnectListAllDomains instead.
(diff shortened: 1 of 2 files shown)


libvirt.go
```diff
@@ -241,6 +241,17 @@ func (l *Libvirt) Disconnected() <-chan struct{} {
 	return l.disconnected
 }
 
+// IsDisconnected indicates whether or not there is still a connection to
```
sam-github, line 244:
```suggestion
// IsDisconnected indicates whether or not there is currently a connection to
```

I'm hoping this will not explode if called before the first connection.

libvirt_test.go
```diff
@@ -832,3 +832,26 @@ func TestConnectGetAllDomainStats(t *testing.T) {
 		t.Fatalf("unexpected typed param value %v", stats[1].Params[1].Value)
 	}
 }
+
+func TestIsDisconnected(t *testing.T) {
+	dialer := libvirttest.New()
+	l := NewWithDialer(dialer)
+
```
sam-github, line 839:
```suggestion
if !l.IsDisconnected() {
		t.Error("expected IsDisconnected == true, got false")
	}
```

sam-github (approved):
lgtm, but I suggest ensuring it is callable both before and after Connect

sam-github:
Also, I wonder if this is better written `IsConnected()`? (with opposite boolean return, of course).

benlemasurier:
I agree with @sam-github, this might be better as `IsConnected()`

connorkuehl:
Thanks @sam-github, @benlemasurier! I've changed it.

sam-github (approved):

PR #179, opened 2022-11-09
SubscribeEvents Now Works

### LINKED ISSUE #154: Inability to subscribe `metdata-change`, `device-added` and `device-removed` events
I am trying to subscribe to these events (in the title) like following:
```
events, err := l.SubscribeEvents(ctx, libvirt.DomainEventIDDeviceAdded, []libvirt.Domain{})
if err != nil {
	log.Fatal(err)
}
```
and then reading from that channel, but there no event comes no matter what I do..

I can see all these in `virsh event --all --loop`:
```
event 'lifecycle' for domain instance-000005ad: Defined Added
event 'lifecycle' for domain instance-000005ad: Started Booted
event 'lifecycle' for domain instance-000005ad: Suspended Paused
event 'lifecycle' for domain instance-000005ad: Resumed Unpaused
event 'device-added' for domain instance-000005ad: net1
event 'metdata-change' for domain instance-000005ad: element http://openstack.org/xmlns/libvirt/nova/1.1
```
but only lifecycle events work fine and can be subscribe/read from:

```
events, err := l.SubscribeEvents(ctx, libvirt.DomainEventIDLifecycle, []libvirt.Domain{})
if err != nil {
	log.Fatal(err)
}
```

any chance I am just missing something? Btw trying to use this on a openstack host with libvirt 5:
```
# virsh version
Compiled against library: libvirt 5.0.0
Using library: libvirt 5.0.0
Using API: QEMU 5.0.0
Running hypervisor: QEMU 2.11.1
```

kat-co:
Previously, the RPC server didn't recognize callback types and so wouldn't unmarshal the responses.

fixes #154

### DIFF
--- a/libvirt.go
+++ b/libvirt.go
@@ -101,6 +101,116 @@ func (m DomainEventCallbackLifecycleMsg) GetCallbackID() int32 {
 	return m.CallbackID
 }
 
+// GetCallbackID returns the callback ID.
+func (e *DomainEventCallbackRebootMsg) GetCallbackID() int32 {
+	return e.CallbackID
+}
+
+// GetCallbackID returns the callback ID.
+func (e *DomainEventCallbackRtcChangeMsg) GetCallbackID() int32 {
+	return e.CallbackID
+}
+
+// GetCallbackID returns the callback ID.
+func (e *DomainEventCallbackWatchdogMsg) GetCallbackID() int32 {
+	return e.CallbackID
+}
+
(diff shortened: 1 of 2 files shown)


connorkuehl (approved):
Great find :slightly_smiling_face:

sam-github:
Is this why Kelly was finding the subscription mechanism not working, and switched to polling?

PR #190, opened 2023-04-12
Fix deadlock if Stream Shutdown races with Push.

multippt:
There's a deadlock scenario if the stream is being shutdown while still receiving events, which can occur if using LifecycleEvents and cancelling the context.

Example reproduction case
```
func TestStreamShutdown(t *testing.T) {
	s := NewStream(1, 2)
	events := testEvents(1)
	s.Shutdown()
	s.Push(events[0])
}
```

### DIFF
--- a/internal/event/stream.go
+++ b/internal/event/stream.go
@@ -41,6 +41,7 @@ type Stream struct {
 	in, out chan Event
 
 	// terminates processing
+	ctx      context.Context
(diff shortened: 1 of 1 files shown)


mpontillo:
Thank you for the contribution! Apologies that it took us so long to incorporate it. We're going to supersede this with #292 and credit you accordingly.

PR #208, opened 2024-02-09
c-for-go: use digitalocean-labs fork

mpontillo:
Te current c-for-go project needs enhancement in order to accomodate generating bindings from the latest libvirt source.

This situation can be addressed by using a private fork (at least, until any required changes can be merged upstream). This commit streamlines the build process for c-for-go to allow building a private version to the bin/ directory. This allows go-libvirt to operate without contaminating the GOPATH with a different version.

This commit adds a 'replace' directive to go.mod in order to allow for the eventual usage of a custom version, but does not yet move utilize a different version (because the commit referenced in go.mod is unchanged).

Because the 'go install' tool recently changed behavior to NOT update the go.mod file, this commit also adds a reference to c-for-go in libvirt.go in order for it to persist, so that the 'replace' directive remains effective.


.github/workflows/main.yml
```diff
@@ -13,19 +13,19 @@ jobs:
...
-      - uses: actions/setup-go@v2
+      - uses: actions/checkout@v4 # respository path: $GITHUB_WORKSPACE
+      - uses: actions/setup-go@v5
         with:
-          go-version: 1.16.2
+          go-version: 1.21.7
 
       - name: Install golint
-        run: go get -u golang.org/x/lint/golint
+        run: go install golang.org/x/lint/golint@latest
```
trapgate, line 26: You could use golangci for this - it'd be a superset of golint, and there's a public action for it.

https://github.com/golangci/golangci-lint

go.mod
```diff
@@ -1,8 +1,28 @@
 module github.com/digitalocean/go-libvirt
 
-go 1.15
+go 1.21
 
 require (
-	github.com/stretchr/testify v1.8.1
-	golang.org/x/tools v0.1.12
+	github.com/stretchr/testify v1.6.1
+	github.com/xlab/c-for-go v0.0.0-00010101000000-000000000000
```
trapgate, line 7: This won't actually work without the replace directive, and that matters because anyone importing this project into their own project isn't going to get the replace directive from this file. In other words, this project will be broken in any package that imports it.

For this to work, you need to switch the dependency over to point at the forked project, and drop the `replace` directive. You could also vendor the forked c-for-go package, but that only makes sense to do when you want to avoid having a public forked package in our repos.

trapgate (commented):
This isn't going to work as-is.

trapgate:
I should also mention that the c-for-go project has accepted patches from us in the past, though I haven't looked at the changes this time round so I don't know if this is something they're likely to object to.

go.mod
mpontillo: Thanks for the feedback; that's unfortunate. I was hoping to avoid changing the module (and all the of the module imports) in `c-for-go`, but I'll go ahead and use that approach for now. Then when I'm ready to contribute the `c-for-go` changes I'll just do so on a rebased branch.

mpontillo:
Now that our `c-for-go` changes are upstream, I'm closing this PR in favor of a new PR to make the compatability changes, and regenerate the bindings against the [`v8.0.0` tag](https://github.com/libvirt/libvirt/tree/v8.0.0) in the `libvirt` tree.

mpontillo:
Superseded by #212.

Issue #89, opened 2019-08-12
TLS connections hang after PR #86


a2geek:
Uncertain exactly why this occurs, but between the 6/26 commit and 7/15 commit (PR #86) the TLS connectivity no longer works. Commit hashes are shown in the `go.mod` file below if that assists.

Please let me know if I need to alter behavior of the code. For now, I'll lock into the 6/26 code.

My environment:
```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.2 LTS"
$ virsh version
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
Using API: QEMU 4.0.0
Running hypervisor: QEMU 2.11.1
```

Bad run sample output (the hang lasts for hours - we left town today and it was still hanging when I returned):
```
$ go run main.go
TCP connection to 'odin:16514' established
Libvirt client created
^Csignal: interrupt
```

Good run sample output:
```
$ go run main.go
TCP connection to 'odin:16514' established
Libvirt client created
Connected to libvirt
Version: 4.0.0
ID	Name		UUID
--------------------------------------------------------
17	vm-852ea340-3a46-471e-7b33-16b30df9bcba	852ea3403a46471e7b3316b30df9bcba
```

`go.mod` file:
```
module tls

go 1.12

// GOOD
require github.com/digitalocean/go-libvirt v0.0.0-20190626172931-4d226dd6c437

// BAD
//require github.com/digitalocean/go-libvirt v0.0.0-20190715144809-7b622097a793
```

`main.go` sample code (sorry, a bit long and hard-coded to locations on my filesystem):
```
package main

import (
	"crypto/tls"
	"crypto/x509"
	"fmt"
	"io/ioutil"
	"log"

	libvirt "github.com/digitalocean/go-libvirt"
)

const connectstring = "odin:16514"
const keyfile = "/home/rob/libvirt-cert/clientkey.pem"
const certfile = "/home/rob/libvirt-cert/clientcert.pem"
const cafile = "/home/rob/libvirt-cert/cacert.pem"

func main() {
	keyfilexml, err := ioutil.ReadFile(keyfile)
	if err != nil {
		panic(err)
	}

	certfilexml, err := ioutil.ReadFile(certfile)
	if err != nil {
		panic(err)
	}

[...]

PR #209, opened 2024-02-19
Add TLS and SSH Dialers, and support libvirt URI parsing

### LINKED ISSUE #139: Support for connection URIs
As part of integrating go-libvirt into terraform-provider-libvirt, I [have implemented](https://github.com/dmacvicar/terraform-provider-libvirt/pull/813/commits/1a86985cac99cc623a0a123f36df475f9b861991) support for connection URIs, including the tcp, tls, unix and ssh transports.

I think this could make go-libvirt easier to use for those using the original bindings, and probably belongs outside of the terraform provider.

Is there interest in having this functionality in go-libvirt?. If that is the case, I could merge the type with [go-libvirt](https://github.com/digitalocean/go-libvirt/blob/master/libvirt.go#L44) and work on a patch.

Looking forward for your feedback. //cc @trapgate

### LINKED ISSUE #143: Tunnel via SSH?
Since #140 was implemented and `New()` deprecated, how is one supposed to do a `qemu+ssh` connection?

I was under the impression this should work:

```
package main

import (
	"fmt"

	"github.com/digitalocean/go-libvirt"
	`github.com/digitalocean/go-libvirt/socket`
	`github.com/digitalocean/go-libvirt/socket/dialers`
)

func main() {

	var conn *libvirt.Libvirt
	var dialer socket.Dialer
	var err error

	dialer = dialers.NewRemote("<IP_ADDRESS_HERE>", dialers.UsePort("22"))

	conn = libvirt.NewWithDialer(dialer)
	if err = conn.ConnectToURI("qemu+ssh://user@<IP_ADDRESS_HERE>/system"); err != nil {
		fmt.Println(err)
		return
	}
}
```

But it just hangs.

I find it hard to believe SSH tunneling was broken by oversight considering it's treated as a first-class connection URI upstream.

### LINKED ISSUE #189: Connecting to Libvirt
Can someone explain to me how you are supposed to actually connect to the libvirt socket in practice?

I am very confused by the API here, and all the examples in the repository use the old `libvirt.New` interface. From what I understand, you have to provide a dialer, but it's not a standard dialer, but there's no way to abstractly construct a dialer from a URI, so you need to already have a parsed URi in order to construct the dialer. Then there is a `ConnectToURI` method, but that still seems to use/require the dialer that is passed to the constructor, so it seems completely useless.

So, I guess I'm just confused what the purpose of the `ConnectToURI` method is if you already have to parse the URI yourself and produce a dialer capable of communicating with the socket before even calling the method.

Long story short, I want to take a URI  as an option to my application just like `virsh --connect qemu:///system` and then connect using the URI. How do I abstractly handle any libvirt-supported URIs? Do I need to manually parse those, and individually support all the connection options myself (ssh, TCP, UDS, etc.)? That is possible, but seems like a ton of error-prone duplication of effort since literally every single `libvirt` client would need to implement this if they wanted anything more complicated than a simple UDS or unencrypted TCP connection.

Maybe I'm missing something, but at the very least, an up-to-date example in the documentation which succinctly shows how to start with a libvirt connection URI and end with a connected instance would be much appreciated. Thanks!

#139 seems related, but has seemingly been dead for almost 2 years.

zaneb:
Add a Dialer for TLS connections. This removes the need for consumers of the library to write a lot of TLS setup code, and abstracts the weird part of the connection protocol discussed in #89.

Add a Dialer for SSH connections, using the golang stdlib ssh library.

Introduce a function to connect to an arbitrary [libvirt connection URI](https://libvirt.org/uri.html). This removes the need for every consumer to reimplement this functionality if they want user-configurable connection parameters, and greatly simplifies the process of opening a connection. The transports that are implemented are `unix`, `tcp`, `tls`, `libssh2`, `libssh`, and `ssh`.

Note that for now all 3 of the ssh transports are implemented using the same golang ssh library for connection, so the behaviour may not always be identical to the libvirt client (in particular the `ssh` transport would normally be expected to pick up settings from the user's `.ssh/config` file by virtue of it shelling out to the ssh client).

Fixes #139
Fixes #143
Fixes #189


trapgate (approved):
Thanks for this submission! go-libvirt has needed better TLS support for a long time.

trapgate:
Hang in there. The CI setup for go-libvirt has gotten a little long-in-the-tooth and needed updating. I've got a separate PR open to fix that; once it has merged you can update this PR and the tests should succeed.

trapgate:
[The CI updates](https://github.com/digitalocean/go-libvirt/pull/210) are merged; if you pull the latest changes into your fork the tests should pass.

zaneb:
Thanks! Rebased the branch on that CI change.

PR #211, opened 2024-02-24
Update README.md, make example current and buildable

yunginnanet:
```
./main.go:6:2: "net" imported and not used
./main.go:8:2: "time" imported and not used
```

additionally, replace deprecated functions with current alternatives.


trapgate (approved):
Thanks for cleaning this up!

PR #212, opened 2024-02-27
Update bindings to libvirt 8

mpontillo:
Now that [our upstream changes have landed in `c-for-go`](https://github.com/xlab/c-for-go/pull/164), `go-libvirt`'s libvirt bindings can be updated to support libvirt `v8.0.0`.

Note that we are roughly following the Ubuntu LTS release cycle with regard to the `libvirt` version supported; I plan to update to test with `v10.0.0` next.


const.gen.go
```diff
@@ -147,357 +73,445 @@ const (
...
+	// PerfParamCacheReferences as defined in libvirt/libvirt-domain.h:2286
 	PerfParamCacheReferences = "cache_references"
-	// PerfParamInstructions as defined in libvirt/libvirt-domain.h:2302
+	// PerfParamInstructions as defined in libvirt/libvirt-domain.h:2296
 	PerfParamInstructions = "instructions"
-	// PerfParamCPUCycles as defined in libvirt/libvirt-domain.h:2312
+	// PerfParamCPUCycles as defined in libvirt/libvirt-domain.h:2306
 	PerfParamCPUCycles = "cpu_cycles"
-	// PerfParamBranchInstructions as defined in libvirt/libvirt-domain.h:2322
+	// PerfParamBranchInstructions as defined in libvirt/libvirt-domain.h:2316
```
sam-github, line 184: We might have no control over this, but I've mixed feelings about the line numbers. On the one hand, could be useful, and they are auto-generated, on the other hand, it makes it really hard to review the generated code to see if its purely additive. I wonder if it would be possible to have the comment have only stable info, file path, and perhaps the name of the symbol the value is pulled from, if the "C" symbol is not exactly `PerfParamBranchInstructions` ?

mpontillo: I like this suggestion. I could look into hacking `c-for-go` to make this configurable.

trapgate (commented):
I think the gen-consts script can be simplified somewhat, and `install-c-for-go.sh` can then be dropped.

Here's my suggested version of `gen-consts.sh`:

```
#!/bin/bash
#
# This runs the first code generator used by go-libvirt: c-for-go. This script
# is run from the 'go generate ./...' command, and only needs to be run when
# changing to a different version of libvirt.

# Set TOP to the root of this repo, and SCRIPTS to the absolute path to the 
# scripts/ directory.
TOP="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
SCRIPTS="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

if [ -z "${LIBVIRT_SOURCE}" ]; then
    echo "Set LIBVIRT_SOURCE to the root of the libvirt sources you want to use first."
    exit 1
fi
if [ -z "$GOBIN" ]; then
    export GOBIN=$TOP/bin
fi

# Make sure c-for-go is installed
echo "Attempting to install c-for-go..."
if ! go install github.com/xlab/c-for-go@v1.1 ; then
    echo "failed to install c-for-go."
    exit 1
fi

# Make sure goyacc is installed (needed for the lvgen/ generator)
echo "Attempting to install goyacc..."
if ! go install golang.org/x/tools/cmd/goyacc@v0.18.0; then
    echo "failed to install goyacc. Please install it manually from https://golang.org/x/tools/cmd/goyacc"
    exit 1
fi

# Ensure fresh output is in libvirt/
rm -rf libvirt/

# Temporarily symlink the libvirt sources to a subdirectory because c-for-go
# lacks a mechanism for us to pass it a search path for header files.
LVDIR=lv_source
ln -sF ${LIBVIRT_SOURCE} ${LVDIR}
if ! bin/c-for-go -nostamp -nocgo -ccincl libvirt.yml; then
    echo "c-for-go failed"
    exit 1
fi
mv libvirt/const.go ${SCRIPTS}/../const.gen.go
rm ${LVDIR}
```

trapgate:
You can drop the `install-c-for-go.sh` script now, I think.

PR #291, opened 2026-06-09
add bidirectional console RPC support

mpontillo:
Because the `DomainOpenConsole()` RPC is only annotated only with `@readstream`, `go-libvirt` does not generate a an `io.Writer` for callers to use. Add a special case which results in a `DomainOpenConsoleBidirectional` to also be written.


DulioCagg (approved):
Honestly I don't understand very well this code, but it LGTM

connorkuehl (approved):

